PR #2, opened 2024-06-26
Get rust2go integration working with mozilla gradle plugin

hellais:
This adds support for adding a layer which calls go code from rust and makes use of the mozilla gradle plugin to build the rust code.


hellais:
It's now building, but it's not copying the .so to the correct directory, which will likely require digging a bit into these settings of the mozilla gradle plugin: https://github.com/mozilla/rust-android-gradle?tab=readme-ov-file#targetdirectory and aligning it to the custom [build.rs](http://build.rs/) for rust2go: https://github.com/ooni/probe-multiplatform/blob/rust/probeengine/build.rs#L51

sdsantos:
The branch will stay as a proof of concept, but we can close the PR.

Issue #6, opened 2024-07-26
Multi-platform tooling: set up logging 


jbonisteel:
create and set up logging as part of tooling for multi-platform app

PR #11, opened 2024-07-29
Tooling: Logging & Platform Info

sdsantos:
Closes https://github.com/ooni/probe-multiplatform/issues/6

Also sets up a `PlatformInfo` interface if we need to use it.


aanorbel (commented):
🐳

aanorbel (approved):
🐳

PR #13, opened 2024-07-29
feat: add support for flavors

aanorbel:
# Proposed changed
- Determine configuration based on flavor.
- Configure application appropriately.
- Add sample text and icons to differentiate branding.

|OONI Probe|News Media Scan|
|-|-|
| Screenshot_20240729_162938 | Screenshot_20240729_163019|


gradle.properties
```diff
@@ -9,4 +9,7 @@ android.useAndroidX=true
 
 #MPP
 kotlin.mpp.androidSourceSetLayoutVersion=2
-kotlin.mpp.enableCInteropCommonization=true
\ No newline at end of file
+kotlin.mpp.enableCInteropCommonization=true
+
+#Flavor
+flavor=dw
```
sdsantos, line 15: This should probably go into the `local.properties` file that does not go into `git`.

composeApp/build.gradle.kts
```diff
@@ -11,6 +11,30 @@ plugins {
     alias(libs.plugins.ktlint)
 }
 
+val flavor: String by project
```
sdsantos, line 14: Should it be `flavor` or something like `organization`? Cause we have other flavors too.

composeApp/build.gradle.kts
```diff
@@ -71,16 +99,24 @@ kotlin {
...
     compileSdk = libs.versions.android.compileSdk.get().toInt()
 
     defaultConfig {
-        applicationId = "org.ooni.probe"
+        applicationId = config.appId
         minSdk = libs.versions.android.minSdk.get().toInt()
         targetSdk = libs.versions.android.targetSdk.get().toInt()
         versionCode = 1
         versionName = "1.0"
+        resValue("string", "app_name", config.appName)
```
sdsantos, line 119: This code is only running for Android isn't it? Plus you are already doing this through the `strings.xml` file.

What we're not covering is having the app name with "Debug" for our debug builds.

The `applicationIdSuffix = ".debug"` config is also only for Android. We might need something similar for iOS?

composeApp/src/commonMain/composeResources/values/strings.xml
```diff
@@ -0,0 +1,2 @@
+<resources>
```
sdsantos, line 1: Are we able to have string resources here as well? Or is the file totally overriden by the ooni and dw variants?

composeApp/src/dwMain/kotlin/org/ooni/probe/ui/Colors.kt
```diff
@@ -0,0 +1,16 @@
...
+val secondaryColor = Color(0xff5db8fe)
+val secondaryLightColor = secondaryColor.copy(alpha = 0.75f)
+val primaryTextColor = Color(0xffffffff)
+val secondaryTextColor = Color(0xff000000)
+val surfaceDark = Color(0xFF161616)
+val surfaceLight = Color(0xFFFFFFFF)
+val backgroundLightColor = Color(0xffF1F0F5)
+val backgroundDarkColor = Color(0xff010100)
+val errorColor = Color(0xFFFF8989)
+val onErrorColor = Color(0xFF000000)
```
sdsantos, line 16: To make sure we're not missing any color on a specific flavor, we might need to wrap them in a class.

composeApp/src/ooniMain/kotlin/Config.kt
```diff
@@ -0,0 +1,3 @@
+object Config {
+    const val ENVIRONMENT: String = "OONI Probe"
```
sdsantos, line 2: Not clear why we need this constant.

sdsantos (approved):
The approach looks solid to me. Left a bunch of questions and comments on things that could be improved.

composeApp/build.gradle.kts
aanorbel: I removed the `app_name` entry from `composeApp/src/androidMain/res/values/strings.xml` which was setting it statically.

Compose strings set in `composeApp/src/dwMain/resources/values/strings.xml` and  `composeApp/src/ooniMain/resources/values/strings.xml` does not influence the manifest file. So i resulted to setting the value here.

Android configuration is by code. For iOS, I have to do custom scheme configuration in Xcode which is not the prettiest thing to do. So i prefer to do it last. We can definitely have a separate debug bundle ID for iOS.

composeApp/src/commonMain/composeResources/values/strings.xml
aanorbel: We can do so right now. I tested locally but it fails on the ci.

gradle.properties
aanorbel: I did not want to load properties from `local.properties` since the file will be expected, so i removed the property and rely on the default or passing a gradle argument.

Issue #7, opened 2024-07-26
Multi-platform tooling: create front-end pattern example 


jbonisteel:
create front-end pattern example (ViewModel)

Issue #9, opened 2024-07-26
Multi-platform tooling: dependency injection


jbonisteel:
set up dependency injection as part of multi-platform tooling

Issue #8, opened 2024-07-26
Multi-platform tooling: navigation


jbonisteel:
create navigation tooling

PR #14, opened 2024-07-29
DI + Navigation + ViewModel

sdsantos:
- Stored dependencies in the App instance of each platform, so they can be reused if we want to
- Setup Jetpack Navigation
- Setup ViewModels with the androidx ViewModel
- Example ViewModel with a single `state` and `onEvent` method approach
- Updated the README to better reflect our goals

All of patterns used are still up for discussion, just went with the approach I'm using in my latest projects, and with the underlying goal of making our code consistent and easily testable.

Closes https://github.com/ooni/probe-multiplatform/issues/7,  https://github.com/ooni/probe-multiplatform/issues/8 and https://github.com/ooni/probe-multiplatform/issues/9

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,75 +1,42 @@
 # OONI Probe Multiplatform
 
-The goal of this repo is to outline a proposed architecture for building a cross platform app that 
-targets Android, iOS and Desktop (windows and macOS).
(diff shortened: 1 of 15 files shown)


aanorbel (approved):
🐳

Issue #15, opened 2024-07-29
Multi-platform tooling: Main Navigation skeleton


jbonisteel:
Create main navigation skeleton for multi-platform app: 

- Dashboard
- Results
- Result Details
- Settings
- Bottom Navigation

PR #21, opened 2024-07-30
Navigation main screens skeleton

sdsantos:
* Implemented a bottom navigation bar
* Added mostly empty screens for results, result and settings
* Specified a pattern for navigation arguments
* Enabled edge-to-edge on Android

Again the patterns specified are up for discussion. The main goal here was to have ViewModels do most of the work. And to avoid that Screens and ViewModels have instances of NavController, to keep them as simple as possible (again, great for testing).

Closes https://github.com/ooni/probe-multiplatform/issues/15


aanorbel (approved):
🐳

PR #27, opened 2024-07-30
Material 3 theming: colors, schemes and custom font

sdsantos:
Colors aren't meant to be final, just a good starting point from the Material Theme Builder from Figma, based on our main colors.


aanorbel (approved):
🐳

Issue #18, opened 2024-07-29
Multi-platform: Implement Engine API


jbonisteel:
Implement engine API for multi-platform application

PR #36, opened 2024-08-05
Finish Engine SDK

sdsantos:
- Don't expose engine settings to the app. They are build internally using few arguments and the preferences.
- Pass the background dispatcher as an argument so it's easier to test the Engine

TODO

- [x] Tests
- [x] Resubmit Results

Closes https://github.com/ooni/probe-multiplatform/issues/18


aanorbel (approved):
🐳

PR #37, opened 2024-08-06
feat: start settings page

### LINKED ISSUE #26: Multi-platform: Setup preferences in database
setup preferences

aanorbel:
Closes #26

## Proposed changes.

-  Add Settings repository and tests.
- Implement template for settings page

Sample UI

|.|.|.|
|-|-|-|
| Screenshot_20240806_101248 | Screenshot_20240806_101304 | Screenshot_20240806_101317 |


composeApp/build.gradle.kts
```diff
@@ -81,6 +81,10 @@ kotlin {
             implementation(libs.bundles.ui)
             implementation(libs.bundles.tooling)
 
+            implementation(libs.androidx.datastore.preferences.core)
+            implementation(libs.androidx.datastore.core.okio)
+            implementation(libs.kotlinx.atomicfu)
```
sdsantos, line 86: I feel we can use more bundles, to avoid having to add here each individual dependency.

composeApp/src/commonMain/kotlin/org/ooni/probe/CreateDataStore.kt
```diff
@@ -0,0 +1,34 @@
...
+private val lock = SynchronizedObject()
+
+/**
+ * Gets the singleton DataStore instance, creating it if necessary.
+ */
+fun getDataStore(
+    producePath: () -> String,
+    migrations: List<DataMigration<Preferences>> = listOf(),
+): DataStore<Preferences> =
+    synchronized(lock) {
```
sdsantos, line 22: We should be calling this just once, I don't think we need to synchronize it.

composeApp/src/commonMain/kotlin/org/ooni/probe/di/Dependencies.kt
```diff
@@ -9,17 +11,21 @@ import org.ooni.engine.NetworkTypeFinder
...
 import org.ooni.probe.ui.results.ResultsViewModel
+import org.ooni.probe.ui.settings.SettingsViewModel
+import org.ooni.probe.ui.settings.category.SettingsCategoryViewModel
 
 class Dependencies(
     val platformInfo: PlatformInfo,
     private val oonimkallBridge: OonimkallBridge,
     private val baseFileDir: String,
     private val cacheDir: String,
+    private val dataStore: DataStore<Preferences>,
```
sdsantos, line 28: It would be nice if this was a factory instead. This way we're creating the DataStore when the app starts up, when we could be creating it just when it really needs to be used. Maybe it will be used soon anyway, but we should delay as much as we can.

Also, we should try to create the datastore in a background coroutine, since it's a disk read/write operation.

composeApp/src/androidUnitTest/kotlin/org/ooni/testing/CreatePreferenceDataStore.kt
```diff
@@ -0,0 +1,14 @@
...
+import android.app.Application
+import androidx.datastore.core.DataStore
+import androidx.datastore.preferences.core.Preferences
+import androidx.test.core.app.ApplicationProvider
+import org.ooni.probe.di.Dependencies
+
+internal actual fun createPreferenceDataStore(): DataStore<Preferences> {
+    val app = ApplicationProvider.getApplicationContext<Application>()
+    return Dependencies.getDataStore(
+        producePath = { app.filesDir.resolve(Dependencies.Companion.DATA_STORE_FILE_NAME).absolutePath },
```
sdsantos, line 12: We should have a separate file name for testing, to avoid corrupting the dev build preferences.

composeApp/src/iosMain/kotlin/org/ooni/probe/SetupDependencies.kt
```diff
@@ -79,3 +84,19 @@ private fun readAssetFile(path: String): String {
 private class BundleMarker : NSObject() {
     companion object : NSObjectMeta()
 }
+
+@OptIn(kotlinx.cinterop.ExperimentalForeignApi::class)
```
sdsantos, line 88: We're using setting our OptIn in the build.gradle file. Do you prefer setting it by file?

composeApp/src/commonTest/kotlin/org/ooni/probe/data/repositories/PreferenceRepositoryTest.kt
```diff
@@ -0,0 +1,117 @@
...
+    fun after() =
+        runBlocking {
+            runTest {
+                preferenceRepository.clear()
+            }
+        }
+
+    @Test
+    fun testAllSettings() =
+        runBlocking {
```
sdsantos, line 32: Usually `runBlocking` isn't needed, `runTest` is enough. Did you run into any trouble just with `runTest`?

sdsantos (changes requested):
The only blocking comments are regarding the shared preference names, which should match the previous app (I know our applicationIds are still different, but we'll need to match them as well in the future).

The bigger architecture suggestions could be made in the future, but I think it's an important discussion to have.

composeApp/src/commonMain/kotlin/org/ooni/probe/data/repositories/PreferenceRepository.kt
```diff
@@ -0,0 +1,154 @@
...
+     */
+    fun getPreferenceKey(
+        name: String,
+        prefix: String? = null,
+        autoRun: Boolean = false,
+    ): String {
+        return "${prefix?.let { "${it}_" } ?: ""}$name${if (autoRun) "_autorun" else ""}"
+    }
+
+    fun allSettings(keys: List<Preferences.Key<*>>): Flow<Map<String, Any?>> =
```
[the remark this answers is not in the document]
aanorbel: I see your point.

composeApp/src/commonMain/kotlin/org/ooni/probe/ui/settings/SettingsScreen.kt
```diff
@@ -1,30 +1,603 @@
...
+            SettingsCategoryItem(
+                title = Res.string.Settings_Advanced_RecentLogs,
+                route = PreferenceCategoryKey.SEE_RECENT_LOGS,
+            )
+        private val webCategory =
+            SettingsCategoryItem(
+                title = Res.string.Settings_Websites_Categories_Label,
+                route = PreferenceCategoryKey.WEBSITES_CATEGORIES,
+                supportingContent = {
+                    Text(stringResource(Res.string.Settings_Websites_Categories_Description))
```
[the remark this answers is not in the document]
aanorbel: Some more work needs to be done on that. I don't want to increase the size of this PR again.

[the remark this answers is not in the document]
sdsantos: Leaving a TODO comment is possible way to keep track of that.

composeApp/src/iosTest/kotlin/org/ooni/testing/CreatePreferenceDataStore.kt
```diff
@@ -0,0 +1,26 @@
...
+
+import androidx.datastore.core.DataStore
+import androidx.datastore.preferences.core.Preferences
+import org.ooni.probe.di.Dependencies
+import platform.Foundation.NSDocumentDirectory
+import platform.Foundation.NSFileManager
+import platform.Foundation.NSURL
+import platform.Foundation.NSUserDomainMask
+
+@OptIn(kotlinx.cinterop.ExperimentalForeignApi::class)
```
sdsantos, line 11: This is probably not needed anymore.

composeApp/src/commonMain/kotlin/org/ooni/probe/ui/settings/category/SettingsCategoryViewModel.kt
```diff
@@ -0,0 +1,72 @@
...
+                preferenceManager.setValueByKey(
+                    key,
+                    it.value,
+                )
+            }
+            _state.update { state ->
+                state.copy(
+                    preference = state.preference?.plus(it.key to it.value),
+                )
+            }
```
sdsantos, line 50: You shouldn't need to do this. The state should update automatically since you subscribed above to the flow from the `preferencesManager.allSettings`.

composeApp/src/commonMain/kotlin/org/ooni/probe/ui/settings/SettingsScreen.kt
```diff
@@ -1,30 +1,604 @@
...
+    open val title: StringResource,
+    open val icon: DrawableResource? = null,
+    open val type: PreferenceItemType,
+    open val key: SettingsKey,
+    open val supportingContent:
+        @Composable()
+        (() -> Unit)? = null,
+)
+
+data class SettingsItem(
```
sdsantos, line 200: I feel this classes should be in separate files, especially because they are not just related with this screen.

composeApp/src/commonMain/kotlin/org/ooni/probe/data/repositories/PreferenceRepository.kt
```diff
@@ -0,0 +1,179 @@
...
+     */
+    fun getPreferenceKey(
+        name: String,
+        prefix: String? = null,
+        autoRun: Boolean = false,
+    ): String {
+        return "${prefix?.let { "${it}_" } ?: ""}$name${if (autoRun) "_autorun" else ""}"
+    }
+
+    fun preferenceKeyFromSettingsKey(
```
sdsantos, line 38: If `getValueByKey` and `setValueByKey` would use `SettingsKey` as well, this method could be private.

sdsantos (approved):

sdsantos (approved):

Issue #25, opened 2024-07-30
Multi-platform: Create and setup database


jbonisteel:
Create and setup database for multi-platform application

PR #38, opened 2024-08-06
Configure database models and repositories

sdsantos:
Closes https://github.com/ooni/probe-multiplatform/issues/25

Notes:
- The `CREATE TABLE` calls where copied straight from the database
- I added the Android sqlite database to the project, with its version number `5.db`. That allows sqldelight to validate our schema and validate future migrations
- Sqldelight is not flexible in terms of table and column names. So instead of worrying about using type converters and `ALTER TABLE`s, I created our own version of each table with the suffix `Model` so we can have our nice types there, and convert everything with our own code. The downside is more boilerplate code, but I feel it's more flexible.
- So the goal is we only touch sqldelight models inside repositories, and only expose our own Models.
- I wrapped all ids with a `Id` class. I feel like it's a good way to avoid mixing up ids, when they all have the same Long type.
- The repositories just have the 2 basic methods list and insert. But I only added those to write tests to ensure everything is working. We can remove those and write whatever we need.
- I created test factories to simplify our work writing tests in the future. Most fields are defaulting to `null` but fell free to change them to more meaningful defaults.
- Almost all fields in the database were nullables, so I kept most of that. But I left some examples of how we can convert some of those nullables into non-null, inside the `toModel` method. We can basically ignore invalid records from the database.
- We probably want to add some indexes in the future, to speed things up


composeApp/src/commonMain/sqldelight/5.db
aanorbel, line 1: why 5

aanorbel (approved):
🐳

composeApp/src/commonMain/sqldelight/5.db
sdsantos: That's the current version of our database in Android. See `Verifying Migrations` at https://cashapp.github.io/sqldelight/2.0.0-alpha05/multiplatform_sqlite/migrations/

PR #39, opened 2024-08-06
NetworkTypeFinder for Android

sdsantos:
Closes https://github.com/ooni/probe-multiplatform/issues/32

Also wrapped our engine calls with our own exception.


aanorbel (approved):
🐳

PR #74, opened 2024-08-20
feat: about page

aanorbel:
## Proposed Changes
- Add Settings page.
- Add support for launching links in the phone browser.

|.|.|
|-|-|
| Screenshot_20240821_105218 |  Screenshot_20240821_105313 |


composeApp/src/commonMain/kotlin/org/ooni/probe/ui/navigation/Navigation.kt
```diff
@@ -56,6 +57,9 @@ fun Navigation(
                         goToSettingsForCategory = {
                             navController.navigate(Screen.SettingsCategory(it).route)
                         },
+                        sendSupportEmail = {
+                            // TODO(norbel): Send support email
```
sdsantos, line 61: Maybe we can use the `launchUrl` function as well, but with `mailto:email@example.org`?

composeApp/src/commonMain/kotlin/org/ooni/probe/ui/navigation/Navigation.kt
```diff
@@ -99,8 +103,14 @@ fun Navigation(
         ) { entry ->
             val category = entry.arguments?.getString("category") ?: return@composable
             when (category) {
-                PreferenceCategoryKey.SEND_EMAIL.name -> {
-                    // TODO: Implement based on platform
+                PreferenceCategoryKey.ABOUT_OONI.name -> {
+                    val viewModel =
```
sdsantos, line 107: In these "dumb" views like the About, where no state is read, I'm OK with not having a ViewModel. But it's your call 😉

composeApp/src/ooniMain/composeResources/values/strings-organization.xml
```diff
@@ -19,4 +19,11 @@
     <string name="performance_datausage" translatable="false">5 - 200 MB</string>
 
     <string name="Settings_About_Label">About OONI</string>
+    <string name="Settings_About_Content_Paragraph">The Open Observatory of Network Interference (OONI) is a free software project under The Tor Project that aims to increase transparency of internet censorship around the world.\n\nSince 2012, OONI\'s global community has been measuring networks in more than 200 countries. Some of these measurements serve as evidence of internet censorship.</string>
```
sdsantos, line 22: The backspace in `OONI\'s` is showing up in the app. If we switch to the curly version `’` there's no need to escape it.

sdsantos (approved):

composeApp/src/ooniMain/composeResources/values/strings-organization.xml
aanorbel: This is markdown actually. I have to update the view to show the markdown properly.

composeApp/src/commonMain/kotlin/org/ooni/probe/ui/navigation/Navigation.kt
aanorbel: Let me update that. I was thinking of a way to add the mail template ( version, device model and other info)

sdsantos: You can add `cc, bcc, subject and body` as query parameters in the url.

composeApp/src/ooniMain/composeResources/values/strings-organization.xml
sdsantos: Ah ok, got it. The curly version would work anyway.

aanorbel: https://github.com/ooni/probe-multiplatform/pull/81

composeApp/src/commonMain/kotlin/org/ooni/probe/ui/navigation/Navigation.kt
aanorbel: I will leave the `viewModel` provides an organized way of event handling from the view

PR #76, opened 2024-08-21
Support for localized month names

sdsantos:


sdsantos:
> I am guessing [this](https://github.com/ooni/probe-android/blob/c82bad7f75b7f6ee512fbdd6a15e8cc555449c52/app/src/main/java/org/openobservatory/ooniprobe/item/DateItem.java#L22) (SimpleDateFormat with Locale) is no longer possible

Nop, no Java dependencies now

PR #84, opened 2024-09-03
feat: add translation files

aanorbel:
## Proposed changes
- Generate localization files from localization repo.
- Add config to show locale preferences in android settings

|.|.|.|
|-|-|-|
| Screenshot_20240903_123248 | Screenshot_20240903_123400 | Screenshot_20240903_123435 |


composeApp/src/androidMain/kotlin/org/ooni/probe/AndroidApplication.kt
```diff
@@ -81,6 +84,17 @@ class AndroidApplication : Application() {
...
 
+    override fun onCreate() {
+        super.onCreate()
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.UPSIDE_DOWN_CAKE) {
+            val localeManager = applicationContext
+                .getSystemService(LocaleManager::class.java)
+            localeManager.overrideLocaleConfig = LocaleConfig(
+                LocaleList.forLanguageTags(getString(R.string.supported_languages)),
+            )
+        }
```
sdsantos, line 95: Why do we need to set the supported languages here as well? Isn't the code on `build.gradle` enough?
```
        resourceConfigurations += config.supportedLanguages
```

sdsantos (approved):

composeApp/src/androidMain/kotlin/org/ooni/probe/AndroidApplication.kt
aanorbel: I checked it out and it doesnt actually enable the languages in the settings.

PR #89, opened 2024-09-04
Run tests screen

### LINKED ISSUE #58: Test: Run tests screen

sdsantos:
Closes #58 

https://github.com/user-attachments/assets/62ba653f-3058-4bd5-97b8-ee6d05599ad2


composeApp/src/commonMain/kotlin/org/ooni/probe/ui/run/RunScreen.kt
```diff
@@ -0,0 +1,238 @@
...
+                                },
+                                onChecked = {
+                                    onEvent(RunViewModel.Event.DescriptorChecked(descriptor, it))
+                                },
+                            )
+                        }
+
+                        if (!descriptorItem.isExpanded) return@forEach
+
+                        items(testItems, key = { it.item.test.name }) { testItem ->
```
aanorbel, line 128: Cannot use name as `key` here. Test names can be repeated, already the case with `NMS` failing with 

```log
java.lang.IllegalArgumentException: Key "web_connectivity" was already used. If you are using LazyColumn/Row please make sure you provide a unique key for each item.
```

My suggestion is to use. name for tests not part of a descriptor and `descriptor_id_${test_name}` when possible.

aanorbel: I think we should make provisions for customizing the items based on the flavor, `NMS` displays the list of inputs instead of the test names here.

Also the items dont present with a checkbox

aanorbel (approved):
🐳

composeApp/src/commonMain/kotlin/org/ooni/probe/ui/run/RunScreen.kt
aanorbel: Additionally, the checkboxes are now presented on the left. See https://m3.material.io/components/checkbox/guidelines

aanorbel (approved):
🐳  pending lint correction

PR #101, opened 2024-09-10
chore: update link installation

aanorbel:
|.|.|
|-|-|
| Screenshot_20240911_114222 | Screenshot_20240911_114316 |

### DIFF
--- a/composeApp/src/commonMain/kotlin/org/ooni/probe/data/models/InstalledTestDescriptorModel.kt
+++ b/composeApp/src/commonMain/kotlin/org/ooni/probe/data/models/InstalledTestDescriptorModel.kt
@@ -23,7 +23,7 @@ data class InstalledTestDescriptorModel(
     val expirationDate: LocalDateTime?,
     val dateCreated: LocalDateTime?,
     val dateUpdated: LocalDateTime?,
-    val revision: String?,
+    val revision: List<String>? = emptyList(),
     val autoUpdate: Boolean,
 ) {
     @Serializable
--- a/composeApp/src/commonMain/kotlin/org/ooni/probe/data/models/InstalledTestDescriptorModel.kt
+++ b/composeApp/src/commonMain/kotlin/org/ooni/probe/data/models/InstalledTestDescriptorModel.kt
@@ -23,7 +23,7 @@ data class InstalledTestDescriptorModel(
     val expirationDate: LocalDateTime?,
     val dateCreated: LocalDateTime?,
     val dateUpdated: LocalDateTime?,
-    val revision: List<String>? = emptyList(),
+    val revisions: List<String>? = emptyList(),
     val autoUpdate: Boolean,
 ) {
     @Serializable
--- a/composeApp/src/commonMain/kotlin/org/ooni/probe/data/models/SnackBarMessage.kt
+++ b/composeApp/src/commonMain/kotlin/org/ooni/probe/data/models/SnackBarMessage.kt
@@ -1,0 +1,9 @@
+package org.ooni.probe.data.models
+
+sealed interface SnackBarMessage {
+    data object AddDescriptorFailed : SnackBarMessage
+
+    data object AddDescriptorCancel : SnackBarMessage
+
+    data object AddDescriptorSuccess : SnackBarMessage
+}
--- a/composeApp/src/commonMain/kotlin/org/ooni/probe/data/models/SnackBarMessage.kt
+++ b/composeApp/src/commonMain/kotlin/org/ooni/probe/data/models/SnackBarMessage.kt
@@ -1,9 +1,0 @@
-package org.ooni.probe.data.models
-
-sealed interface SnackBarMessage {
-    data object AddDescriptorFailed : SnackBarMessage
-
-    data object AddDescriptorCancel : SnackBarMessage
-
-    data object AddDescriptorSuccess : SnackBarMessage
-}
--- a/composeApp/src/commonMain/kotlin/org/ooni/probe/domain/DeleteTestDescriptor.kt
+++ b/composeApp/src/commonMain/kotlin/org/ooni/probe/domain/DeleteTestDescriptor.kt
@@ -1,0 +1,17 @@
+package org.ooni.probe.domain
+
+import org.ooni.probe.data.models.InstalledTestDescriptorModel
+import org.ooni.probe.data.models.toDescriptor
+import org.ooni.probe.data.repositories.PreferenceRepository
+
+class DeleteTestDescriptor(
+    private val preferencesRepository: PreferenceRepository,
+    private val deleteByRunId: suspend (InstalledTestDescriptorModel.Id) -> Unit,
+) {
+    suspend operator fun invoke(testDescriptor: InstalledTestDescriptorModel) {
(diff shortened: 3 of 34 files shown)


composeApp/src/commonMain/kotlin/org/ooni/probe/ui/shared/NotificationMessages.kt
```diff
@@ -0,0 +1,33 @@
...
+import org.ooni.probe.LocalSnackbarHostState
+import org.ooni.probe.data.models.SnackBarMessage
+
+@Composable
+fun NotificationMessages(
+    message: List<SnackBarMessage>,
+    onMessageDisplayed: (SnackBarMessage) -> Unit = { },
+) {
+    val snackbarHostState = LocalSnackbarHostState.current ?: return
+    val errorMessage = when (message.firstOrNull()) {
```
sdsantos, line 20: It's not always an errorMessage right?

composeApp/src/commonMain/kotlin/org/ooni/probe/ui/shared/NotificationMessages.kt
```diff
@@ -0,0 +1,33 @@
...
+import ooniprobe.composeapp.generated.resources.LoadingScreen_Runv2_Failure
+import ooniprobe.composeapp.generated.resources.Res
+import org.jetbrains.compose.resources.stringResource
+import org.ooni.probe.LocalSnackbarHostState
+import org.ooni.probe.data.models.SnackBarMessage
+
+@Composable
+fun NotificationMessages(
+    message: List<SnackBarMessage>,
+    onMessageDisplayed: (SnackBarMessage) -> Unit = { },
```
sdsantos, line 17: This argument is not being used, meaning there's a part missing in the messages queue.

The ViewModel needs to append messages to the list/queue, and when the ViewModel receives the event of a message being displayed, it needs to remove the message from the list/queue.

This is important because otherwise, if the view gets recreated, it will show the last message shown again, because it wasn't taken off the list. And if 2 messages are triggered at the same time from the ViewModel, one will override the other, while in a queue, they are shown in order.

composeApp/src/commonMain/kotlin/org/ooni/probe/ui/shared/NotificationMessages.kt
```diff
@@ -0,0 +1,33 @@
...
+) {
+    val snackbarHostState = LocalSnackbarHostState.current ?: return
+    val errorMessage = when (message.firstOrNull()) {
+        SnackBarMessage.AddDescriptorSuccess -> stringResource(Res.string.AddDescriptor_Toasts_Installed)
```
sdsantos, line 21: If you want to get a string from inside `LaunchedEffect`, which is not a `@Composable`, you can use the `suspend getString(StringResource)` method. Just do you know.

composeApp/src/commonMain/kotlin/org/ooni/probe/data/models/SnackBarMessage.kt
```diff
@@ -0,0 +1,9 @@
+package org.ooni.probe.data.models
+
+sealed interface SnackBarMessage {
```
sdsantos, line 3: This is not any SnackbarMessage, but the AddDescriptor specific messages. Maybe this could have a more specific name, or be inside the AddDescriptorViewModel class?

sdsantos (changes requested):

composeApp/src/commonMain/kotlin/org/ooni/probe/domain/DeleteTestDescriptor.kt
```diff
@@ -0,0 +1,17 @@
...
+
+class DeleteTestDescriptor(
+    private val preferencesRepository: PreferenceRepository,
+    private val deleteByRunId: suspend (InstalledTestDescriptorModel.Id) -> Unit,
+) {
+    suspend operator fun invoke(testDescriptor: InstalledTestDescriptorModel) {
+        preferencesRepository.removeDescriptorPreferences(
+            descriptor = testDescriptor.toDescriptor(),
+        )
+        deleteByRunId(testDescriptor.id)
```
sdsantos, line 15: Shouldn't we be deleting the results/measurements and files related with this descriptor too? 

This way, if the user installs the same link again, they will see the results back. But they will be clogging some space in the meantime. In alternative we can also soft delete the descriptor (hide from the homepage and auto-run, but keep in the DB).

composeApp/src/commonMain/kotlin/org/ooni/probe/data/models/InstalledTestDescriptorModel.kt
```diff
@@ -23,7 +23,7 @@ data class InstalledTestDescriptorModel(
     val expirationDate: LocalDateTime?,
     val dateCreated: LocalDateTime?,
     val dateUpdated: LocalDateTime?,
-    val revision: String?,
+    val revision: List<String>? = emptyList(),
```
sdsantos, line 26: `revisions` then?

composeApp/src/commonMain/kotlin/org/ooni/probe/domain/FetchDescriptor.kt
```diff
@@ -15,12 +17,26 @@ class FetchDescriptor(
...
             "GET",
-            "https://api.dev.ooni.io/api/v2/oonirun/links/$descriptorId",
+            "${OrganizationConfig.ooniApiBaseUrl}/api/v2/oonirun/links/$descriptorId",
             TaskOrigin.OoniRun,
         ).map { result ->
             result?.let {
-                json.decodeFromString<OONIRunDescriptor>(it).toModel()
+                json.decodeFromString<OONIRunDescriptor>(it).toModel().copy(
+                    revision = fetchRevisions(descriptorId).get()?.revisions,
+                )
```
sdsantos, line 26: We probably should try/catch these json decodings

composeApp/src/commonMain/kotlin/org/ooni/probe/ui/descriptor/add/AddDescriptorViewModel.kt
```diff
@@ -41,7 +41,9 @@ class AddDescriptorViewModel(
                     )
                 }
             }.onFailure {
-                onError()
+                Logger.e(it) { "Failed to fetch descriptor" }
+                _state.value = state.value.copy(messages = listOf(SnackBarMessage.AddDescriptorFailed))
```
sdsantos, line 45: You need to append messages to the queue, and not remove all existing messages:
```
_state.update { it.copy(messages = it.message + SnackBarMessage.AddDescriptorFailed) } 
```

sdsantos (approved):

composeApp/src/commonMain/kotlin/org/ooni/probe/domain/DeleteTestDescriptor.kt
aanorbel: We should be deleting associated measurements. I saw you doing some measurements stuff and wanted to call some ready made API's if possible to do so instead of reinventing the wheel

sdsantos (approved):

PR #111, opened 2024-09-12
feat: background run on iOS

aanorbel:
Closes: https://github.com/ooni/probe-multiplatform/issues/88 
Closes: https://github.com/ooni/probe-multiplatform/issues/87

Note : Use `e -l objc -- (void)[[BGTaskScheduler sharedScheduler] _simulateLaunchForTaskWithIdentifier:@"org.ooni.probe.autorun-task"]` to test iOS background task in actual device.

PS: Task will not launch in in emulator 

https://developer.apple.com/documentation/uikit/app_and_environment/scenes/preparing_your_ui_to_run_in_the_background/using_background_tasks_to_update_your_app

https://developer.apple.com/documentation/backgroundtasks


composeApp/src/commonMain/kotlin/org/ooni/probe/domain/GetAutoRunSettings.kt
```diff
@@ -0,0 +1,29 @@
+package org.ooni.probe.domain
+
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.map
+import org.ooni.probe.data.models.AutoRunParameters
+import org.ooni.probe.data.models.SettingsKey
+
+class GetAutoRunSettings(
```
sdsantos, line 8: Since we now have this class, we could use it directly in `ObserveAndConfigureAutoRun`, instead of having the duplicated code there.

composeApp/src/iosMain/kotlin/org/ooni/probe/background/RunOperation.kt
```diff
@@ -0,0 +1,59 @@
...
+) : NSOperation() {
+    init {
+        require(spec != null || getAutoRunSpecification != null) {
+            "Either spec or getAutoRunSpecification must be provided"
+        }
+    }
+
+    override fun main() {
+        Logger.d { "Running operation" }
+        val coroutineScopeScope = CoroutineScope(SupervisorJob() + Dispatchers.Main)
```
sdsantos, line 34: Should this be running on the main thread? I know we switch to the background dispatcher on the engine and DB, but just wondering if we need to specify Main here, or we can just go with the Default.

sdsantos (approved):
Looks good and it runs fine.

But I'll have to trust you on the iOS code, since I'm not familiar with it 😅

composeApp/src/iosMain/kotlin/org/ooni/probe/background/RunOperation.kt
aanorbel: Updated

PR #125, opened 2024-09-20
Various Run Test improvements

sdsantos:
* Update AGP
* Store engine URL
* Remove notification sound and vibration
* Don't run web connectivity test without URLs


aanorbel (approved):
🐳

PR #129, opened 2024-09-25
feat: Configure Sentry crash reporting

### LINKED ISSUE #106: Crash Monitoring with Sentry

sdsantos:
Closes #106

### DIFF
--- a/composeApp/build.gradle.kts
+++ b/composeApp/build.gradle.kts
@@ -88,7 +88,7 @@ kotlin {
     iosSimulatorArm64()
 
     cocoapods {
-        ios.deploymentTarget = "12.0"
+        ios.deploymentTarget = "15.3"
 
         version = "1.0"
         summary = "Compose App"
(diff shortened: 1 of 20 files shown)


aanorbel (approved):
🐳

aanorbel:
Please fix broken test

composeApp/build.gradle.kts
```diff
@@ -88,7 +88,7 @@ kotlin {
     iosSimulatorArm64()
 
     cocoapods {
-        ios.deploymentTarget = "12.0"
+        ios.deploymentTarget = "15.3"
```
aanorbel, line 91: If sentry is the only thing forcing us to jump versions, can we use an older version of sentry?

sdsantos: This was me during bug fixing. But I just copied the minimum we had on the iOS config files. I think we had a lower minimum here than there, so I just matched it. Our minimum is 15.3 right?

aanorbel: its `13.0`

sdsantos: Oh, then I have to fix it on a bunch of different places.

sdsantos: Addressed in https://github.com/ooni/probe-multiplatform/pull/133

PR #135, opened 2024-10-02
fix: Onboarding text cut with large fonts

### LINKED ISSUE #132: Accessibility: text on some screen cuts out when the font size is increased.
See also https://github.com/ooni/probe/issues/2749

sdsantos:
Closes #132


sdsantos:
Will fix on another PR

PR #145, opened 2024-10-03
Results UI

sdsantos:
Screenshot_20241003_183931


sdsantos:
I'll probably still tweak the Result items.

composeApp/src/commonMain/kotlin/org/ooni/probe/ui/results/ResultsScreen.kt
```diff
@@ -85,33 +87,35 @@ fun ResultsScreen(
...
-                list = state.descriptorFilters,
-                onFilterChanged = { onEvent(ResultsViewModel.Event.DescriptorFilterChanged(it)) },
-                modifier = Modifier.weight(3f).padding(horizontal = 4.dp),
-            )
+                DescriptorFilter(
+                    current = state.filter.descriptor,
+                    list = state.descriptorFilters,
+                    onFilterChanged = { onEvent(ResultsViewModel.Event.DescriptorFilterChanged(it)) },
+                    modifier = Modifier.weight(3f).padding(horizontal = 4.dp),
+                )
```
aanorbel, line 110: Title for installed descriptors goes to multiple lines which is ok but then there is no way to tell where one end and the other starts.

Screenshot_20241003_194835

aanorbel (approved):
🐳

PR #147, opened 2024-10-04
fix(workers): worker input error

aanorbel:


sdsantos (approved):

sdsantos:
@aanorbel just need a fix for kotlin lint

aanorbel:
the error has something to do with `Imports must be ordered in lexicographic order without any empty lines in-between with "java", "javax", "kotlin" and aliases in the end` I have run the formatter a couple of times and the error persists.

aanorbel:
The fix seems to be to add `disabled_rules = import-ordering` to the editor config.

PR #150, opened 2024-10-04
Android UI instrumented tests

### LINKED ISSUE #123: Automated Functional Tests for the whole Run flow

sdsantos:
Closes #123

- [x] Onboarding
- [x] Settings
- [x] Configure CI
- [x] Run Test + Results

### DIFF
--- a/composeApp/src/androidInstrumentedTest/kotlin/org/ooni/probe/test/OnboardingTest.kt
+++ b/composeApp/src/androidInstrumentedTest/kotlin/org/ooni/probe/test/OnboardingTest.kt
@@ -1,0 +1,78 @@
+package org.ooni.probe.test
+
+import androidx.compose.ui.test.isDisplayed
+import androidx.compose.ui.test.junit4.createEmptyComposeRule
+import androidx.compose.ui.test.onNodeWithContentDescription
+import androidx.compose.ui.test.onNodeWithText
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import kotlinx.coroutines.flow.first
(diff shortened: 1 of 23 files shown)


composeApp/src/androidInstrumentedTest/kotlin/org/ooni/probe/test/OnboardingTest.kt
```diff
@@ -0,0 +1,63 @@
...
+    @Before
+    fun setUp() {
+        start()
+    }
+
+    @Test
+    fun onboarding() =
+        runTest {
+            compose.waitUntil { compose.onNodeWithText("What is OONI Probe?").isDisplayed() }
+            compose.onNodeWithText("Got It").performClick()
```
aanorbel, line 39: Would this not be better to run these with Tag/Id so we can Once in a while run the tests with different device configuration for language like RTL and others

sdsantos: That's a good point. 

In one hand, testing the actual strings as the users see them, will make sure they are seeing the right content we want. And it's simpler, because we don't have to add test tags to everything. But on the other hand, to test in other languages, we need to write the tests for those languages, with the changed strings.

If we use test tags everywhere, the tests would still pass if we switch languages and text direction. But we wouldn't be testing if the users are actually seeing the right text they should be seeing.

There are upsides and downsides to both methods. What would you feel is the best approach?

aanorbel (approved):
🐳 

We need to fix tests

PR #179, opened 2024-10-15
Protect navigation from duplicated actions and empty screens

### LINKED ISSUE #178: prevent back button from being clicked multiple times.
Clicking the back button multiple times can land you in an error state where nothing is shown to user.

System navigation works fine, but clicking on the back button on any pack multiple times performs the action multiple times.

sdsantos:
Closes #178


aanorbel (approved):
🐳

PR #198, opened 2024-10-21
Disable javascript and other links from webview

### LINKED ISSUE #195: Sandbox Webview
Ensure the webview for the measurements page doesn't leak personal information, or open any other link inside the app.

sdsantos:
Closes #195

### DIFF
--- a/composeApp/src/commonMain/kotlin/org/ooni/probe/ui/measurement/MeasurementScreen.kt
+++ b/composeApp/src/commonMain/kotlin/org/ooni/probe/ui/measurement/MeasurementScreen.kt
@@ -22,8 +22,12 @@ import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.unit.dp
 import co.touchlab.kermit.Logger
+import com.multiplatform.webview.request.RequestInterceptor
+import com.multiplatform.webview.request.WebRequest
+import com.multiplatform.webview.request.WebRequestInterceptResult
 import com.multiplatform.webview.web.LoadingState
 import com.multiplatform.webview.web.WebView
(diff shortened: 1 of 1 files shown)


composeApp/src/commonMain/kotlin/org/ooni/probe/ui/measurement/MeasurementScreen.kt
```diff
@@ -45,7 +49,22 @@ fun MeasurementScreen(
     LaunchedEffect(url) { Logger.i("URL: $url") }
 
     val webViewState = rememberWebViewState(url)
-    val webViewNavigator = rememberWebViewNavigator()
+    webViewState.webSettings.isJavaScriptEnabled = false
```
aanorbel, line 52: By disabling javascript, we lose the ability to view raw measurement data.
Screenshot_20241022_100007

aanorbel (changes requested):
Re-enable javascript.

composeApp/src/commonMain/kotlin/org/ooni/probe/ui/measurement/MeasurementScreen.kt
sdsantos: @aanorbel I noticed, but those pages will eventually have a mobile friendly view, that won't have any link or interaction, according to @hellais. That's why I went with this change.

aanorbel: https://github.com/ooni/explorer/issues/951 this is the issue on explorer

sdsantos: Replied there with a comment 👍

PR #202, opened 2024-10-22
feat: added proxy label to running screen

aanorbel:
Closes: https://github.com/ooni/probe-multiplatform/issues/201

### DIFF
--- a/composeApp/src/commonMain/kotlin/org/ooni/probe/ui/running/RunningViewModel.kt
+++ b/composeApp/src/commonMain/kotlin/org/ooni/probe/ui/running/RunningViewModel.kt
@@ -10,11 +10,10 @@ import kotlinx.coroutines.flow.filterIsInstance
 import kotlinx.coroutines.flow.launchIn
 import kotlinx.coroutines.flow.onEach
 import kotlinx.coroutines.flow.update
+import kotlinx.coroutines.launch
 import org.ooni.probe.data.models.ProxySettings
-import org.ooni.probe.data.models.SettingsKey
 import org.ooni.probe.data.models.TestRunError
 import org.ooni.probe.data.models.TestRunState
-import org.ooni.probe.data.repositories.PreferenceRepository
 
 class RunningViewModel(
     onBack: () -> Unit,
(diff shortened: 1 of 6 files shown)


composeApp/src/commonMain/kotlin/org/ooni/probe/ui/running/RunningViewModel.kt
```diff
@@ -10,23 +10,40 @@ import kotlinx.coroutines.flow.filterIsInstance
...
+                SettingsKey.PROXY_PROTOCOL,
+                SettingsKey.PROXY_HOSTNAME,
+                SettingsKey.PROXY_PORT,
+            ),
+        ).onEach { proxySettings ->
+            val proxy = ProxySettings.newProxySettings(
+                protocol = proxySettings[SettingsKey.PROXY_PROTOCOL] as? String,
+                hostname = proxySettings[SettingsKey.PROXY_HOSTNAME] as? String,
+                port = proxySettings[SettingsKey.PROXY_PORT] as? String,
+            ).getProxyString()
```
sdsantos, line 44: This could be in a GetProxySettings domain class. That would make the code here simpler, and maybe it could be reused on `GetEnginePreferences` for example.

composeApp/src/commonMain/kotlin/org/ooni/probe/ui/running/RunningViewModel.kt
```diff
@@ -10,23 +10,40 @@ import kotlinx.coroutines.flow.filterIsInstance
...
     cancelTestRun: () -> Unit,
+    preferencesRepository: PreferenceRepository,
 ) : ViewModel() {
     private val events = MutableSharedFlow<Event>(extraBufferCapacity = 1)
 
     private val _state = MutableStateFlow(State())
     val state = _state.asStateFlow()
 
     init {
+        preferencesRepository.allSettings(
```
sdsantos, line 33: In theory this doesn't tell us if the actual test is running with a VPN or not. The user could have changed the settings while a test is running. But I think it's good enough for the use case.

PR #214, opened 2024-10-23
Add database indexes

sdsantos:
I believe these will be important for users with a lot of results and measurements, specially since we're now storing auto-runs.

I picked all fields being used for ordering and filtering of Results and Measurements, the tables that will increase the most in size, and are used most often.


aanorbel (approved):
🐳

Issue #208, opened 2024-10-23
Support for F-Droid


jbonisteel:
We want to continue to publish OONI probe on the F droid app store, we need to disable notifications and crash reporting to do so. This issue is to track how we will do that and implementing it so we can support this flavour of the app.

PR #216, opened 2024-10-23
feat: add support for `f-droid` build

aanorbel:
Closes https://github.com/ooni/probe-multiplatform/issues/208


composeApp/src/androidFdroid/kotlin/org/ooni/probe/config/FlavorConfig.kt
```diff
@@ -0,0 +1,10 @@
+package org.ooni.probe.config
+
+class FlavorConfig : FlavorConfigInterface {
+    override val optionalFeatures: Set<OptionalFeature>
+        get() = emptySet()
```
sdsantos, line 5: You only need to override this `val`.

sdsantos (changes requested):
I believe we need to remove the dependencies from APK all together for Fdroid, not just not enable them. I don't think Fdroid will approve our builds if we're including Firebase Cloud Messaging in the future for example...

Issue #230, opened 2024-10-29
NMS iOS: Autoruns not working as expected


jbonisteel:
Note: although I am specifically referencing NMS iOS in this ticket, I assume this same issue would happen with OONI Probe iOS.

This is what I have noticed while testing NMS iOS:

- When I first installed NMS iOS (so totally fresh install on my device) autorun worked as expected **when** the device was plugged in and the 'only while charging' setting was toggled ON. 
- I then toggled OFF the 'only while charging' setting and unplugged my device  
  - Expected results: autoruns should still happen 
  - Actual results: autoruns appeared to stop, no new autoruns happened over multi-hour period
- Then I toggled 'only while charging' back ON and plugged in my device once again-
  - Expected results: autoruns should start up again
  - Actual results: autorun has not started again, even when device is left for long periods (multple hrs). Also worth noting that I deleted and reinstalled the app multiple times and I could still not get autoruns going again, no matter the configuration of settings.  

We had previously discussed that autoruns weren't on the critical path for launching NMS iOS. So we should discuss what that means in terms of the various settings that we have in the app (should we leave them, even though broken, or what). 

Would it at least be possible to have autorun working consistently at least while charging is happening? 

I also assume this same issue would happen on OONI Probe iOS, where autoruns are more critical

aanorbel:
Will look into this. There should be some areas of improvements.

jbonisteel:
>Then I toggled 'only while charging' back ON and plugged in my device once again-
Expected results: autoruns should start up again
Actual results: autorun has not started again, even when device is left for long periods (multple hrs). Also worth noting that I deleted and reinstalled the app multiple times and I could still not get autoruns going again, no matter the configuration of settings.  

I still can't get autorun to re-start as described above. I have tried fully deleting the app and re-installing it as well, and no luck

aanorbel:
Updated the flow on the latest release and should be working as expected.

jbonisteel:
I am still seeing the issue where autoruns do not appear to start again after adjusting the settings. 

I am including some screen recordings. 

In this first recording - you will see the last autorun was 3 hrs ago. Since that last autorun, I disabled the 'only while charging setting' (and phone is unplugged) and no new autoruns have happened for the past three hours. This is consistent with the current issue we see with autoruns. 

https://github.com/user-attachments/assets/10fc93ef-cdd1-4dc6-a3de-1d23940900a8


[...]

PR #234, opened 2024-10-30
chore: update `ios` background scheduling

aanorbel:
Closes https://github.com/ooni/probe-multiplatform/issues/230


sdsantos (approved):
Going to have to trust you on this one. So it looked like the re-configuration wasn't working properly because we didn't always cancelled it before?

aanorbel:
I want to distribute on test flight and test before merging. Background tasks are quite challenging to debug on iOS

PR #237, opened 2024-10-30
Improve checkbox and switch accessibility

sdsantos:
I recently learned that's how you're supposed to implement Checkboxes and Switches with Compose, to maintain properly accessibility.


aanorbel (approved):
🐳

Issue #238, opened 2024-11-01
Add support for sharing a link to the OONI Probe app


aanorbel:
https://github.com/ooni/probe/issues/1997

PR #239, opened 2024-11-01
feat: Add support for sharing a link to the OONI Probe app

aanorbel:
Closes https://github.com/ooni/probe-multiplatform/issues/238


composeApp/src/androidMain/kotlin/org/ooni/probe/MainActivity.kt
```diff
@@ -50,18 +51,29 @@ class MainActivity : ComponentActivity() {
...
-        when (uri.host) {
-            "runv2",
-            OrganizationConfig.ooniRunDomain,
-            -> {
-                val id = uri.lastPathSegment ?: return
-                deepLinkFlow.tryEmit(DeepLink.AddDescriptor(id))
-            }
+        if (intent.action != Intent.ACTION_VIEW && intent.action != Intent.ACTION_SEND) return
+
+        if (intent.action == Intent.ACTION_VIEW) {
```
sdsantos, line 56: A `when` might be simpler now that there are multiple options.

composeApp/src/androidMain/kotlin/org/ooni/probe/MainActivity.kt
```diff
@@ -50,18 +51,29 @@ class MainActivity : ComponentActivity() {
...
+                }
+            }
+        } else if (intent.action == Intent.ACTION_SEND) {
+            val url = intent.getStringExtra(Intent.EXTRA_TEXT) ?: return
+            if (Patterns.WEB_URL.matcher(url).matches()) {
+                deepLinkFlow.tryEmit(DeepLink.RunUrls(url))
+            } else {
+                Logger.e { "Unknown deep link: $url" }
+                return
             }
```
sdsantos, line 77: Unlike the deeplink, I think this case needs a bit more feedback to the user. We can't filter that the intents we get are URLs. That means the OONI app will appear in a lot more Share dialogs. And users won't know what the app supports. So, even if the users shares an invalid URL (a piece of text), we should open the ChooseWebsites screen and show an error explaining That OONI only supports sharing URLs to test. What do you think?

Also, we need to be careful with Logger.e or Logger.w. They always generate a non-crashing error on Sentry.

sdsantos: The same should probably also apply to the iOS app.

composeApp/src/androidMain/AndroidManifest.xml
```diff
@@ -58,6 +58,22 @@
...
+            android:exported="true"
+            android:enabled="@string/ooni_run_enabled"
+            android:name=".ShareRunActivity"
+            android:targetActivity=".MainActivity">
+
+            <intent-filter>
+                <action android:name="android.intent.action.SEND" />
+                <category android:name="android.intent.category.DEFAULT" />
+
+                <action android:name="org.openobservatory.ooniprobe.nettest" />
```
sdsantos, line 71: Why not use `"${applicationId}.nettest"` like on the other intent-filter?

sdsantos (changes requested):

composeApp/src/androidMain/kotlin/org/ooni/probe/MainActivity.kt
```diff
@@ -50,18 +51,31 @@ class MainActivity : ComponentActivity() {
...
-        if (intent.action != Intent.ACTION_VIEW) return
-        val uri = intent.data ?: return
-        when (uri.host) {
-            "runv2",
-            OrganizationConfig.ooniRunDomain,
-            -> {
-                val id = uri.lastPathSegment ?: return
-                deepLinkFlow.tryEmit(DeepLink.AddDescriptor(id))
-            }
+        if (intent.action != Intent.ACTION_VIEW && intent.action != Intent.ACTION_SEND) return
```
sdsantos, line 54: With the `when` we can skip this.

composeApp/src/androidMain/kotlin/org/ooni/probe/MainActivity.kt
```diff
@@ -50,7 +51,23 @@ class MainActivity : ComponentActivity() {
...
+        when (intent.action) {
+            Intent.ACTION_VIEW -> manageOoniRun(intent)
+            Intent.ACTION_SEND -> manageSend(intent)
+            else -> return
+        }
+    }
+
+    private fun manageSend(intent: Intent) {
+        val url = intent.getStringExtra(Intent.EXTRA_TEXT) ?: return
+        if (Patterns.WEB_URL.matcher(url).matches()) {
```
sdsantos, line 63: It would have been better to just let the ChooseWebsites handle the url, and validate if it's a good URL. It uses a different validation logic than `Patterns.WEB_URL`. `Deeplink.Error` is also not a very clear name.

composeApp/src/commonMain/kotlin/org/ooni/probe/App.kt
```diff
@@ -111,6 +114,14 @@ fun App(
                 navController.navigate("add-descriptor/${deepLink.id}")
                 onDeeplinkHandled()
             }
+            is DeepLink.RunUrls -> {
+                navController.navigate(Screen.ChooseWebsites(deepLink.url).route)
+                onDeeplinkHandled()
+            }
+            DeepLink.Error -> {
+                snackbarHostState.showSnackbar(getString(Res.string.LoadingScreen_Runv2_Failure))
```
sdsantos, line 122: The "Error" message isn't clear enough for the user to understand what happened. We should explain OONI Probe only supports URLs to test.

sdsantos (changes requested):

sdsantos (changes requested):
I suggested validating the URL inside ChooseWebsites, but we need to do it before adding it to the Screen. Otherwise a bunch of random text is added to the URL field, and the user is still not clear of what is going on.

I think we should validate the `initialUrl` in the ViewModel, and then show a specific error message before adding the url on the screen, if it's invalid.

composeApp/src/commonMain/kotlin/org/ooni/probe/ui/choosewebsites/ChooseWebsitesViewModel.kt
```diff
@@ -19,10 +19,19 @@ class ChooseWebsitesViewModel(
...
+    initialUrl: String? = null,
 ) : ViewModel() {
     private val events = MutableSharedFlow<Event>(extraBufferCapacity = 1)
 
-    private val _state = MutableStateFlow(State())
+    private val _state = MutableStateFlow(
+        State(
+            websites = listOf(
+                initialUrl?.let {
+                    WebsiteItem(url = it, hasError = it.isValidUrl())
```
sdsantos, line 30: It should be `hasError = !it.isValidUrl()`.

sdsantos (changes requested):

composeApp/src/commonMain/kotlin/org/ooni/probe/ui/choosewebsites/ChooseWebsitesViewModel.kt
aanorbel, line 30:
```suggestion
                    WebsiteItem(url = it, hasError = !it.isValidUrl())
```

sdsantos (approved):

Issue #251, opened 2024-11-11
NMS iOS: Email support doesn't seem to function properly


jbonisteel:
Steps to reproduce:

- In NMS iOS app, navigate to settings>send email to support
- Hit cancel, then delete draft
- The email view is still open, I would have expected it to close
- Hitting either cancel or send doesn't do anything, email view is still stuck open. You have to swipe down to close it
- Also sending doesn't either - when I tap the blue 'send' icon, nothing happens 

https://github.com/user-attachments/assets/e2f2ddca-3431-4d64-8b89-d990ff8468c2

jbonisteel:
Also - realized that it does actually seem like the email does send, however, the email view remains open so it isn't clear to the user that anything has happened

PR #253, opened 2024-11-11
fix: ios email error

aanorbel:
https://github.com/ooni/probe-multiplatform/issues/251


composeApp/src/iosMain/kotlin/org/ooni/probe/background/RunOperation.kt
```diff
@@ -14,7 +16,8 @@ class RunOperation(
 ) : NSOperation() {
     override fun main() {
         Logger.d { "Running operation" }
-        GlobalScope.launch {
+        val coroutineScope = CoroutineScope(SupervisorJob() + Dispatchers.Default)
```
sdsantos, line 19: This doesn't look related with the iOS issue. Trying to fix background auto-run?

sdsantos (approved):

Issue #250, opened 2024-11-11
NMS iOS: Unsure if app continues to run tests in background


jbonisteel:
I am unsure if tests are continuing to run when I put the app in the background. See the attached screen recording.

Steps to reproduce:

- Open NMS and initiate a test
- Navigate to a different app for a while (in the recording, I go check out the news app for a little bit)
- Come back to NMS app and the time hasn't really decreased

So I am not sure if it is just the tests are really slow or if something is happening and they pause while I use a different app. I have tried testing this multiple times with tests that take longer and have left the NMS app in the background for longer with the same results. If someone could also test this and verify if it is working as intended or if, indeed, there is a bug - that would be great!

https://github.com/user-attachments/assets/09ef0b5d-48c3-4411-9ef6-2b5f6be569d6

aanorbel:
I can confirm there is a bug here. The tests seem to be paused till the user resumes the app. Looking into the issue

PR #258, opened 2024-11-11
fix: background runner strategy

aanorbel:
Closes https://github.com/ooni/probe-multiplatform/issues/250 , https://github.com/ooni/probe/issues/2361 , https://github.com/ooni/probe/issues/2175

## Proposed Changes

- Implemented a new `IosBackgroundRunner` class to handle running tasks in the background on iOS devices. The `IosBackgroundRunner` class sets up a background task and an audio player to keep the app running in the background, even when the app is not in the foreground.
- Delete `RunOperation`

https://hayageek.com/ios-long-running-background-task/


iosApp/iosApp/background/IosBackgroundRunner.swift
```diff
@@ -0,0 +1,70 @@
...
+import AVFAudio
+import UIKit
+
+class IosBackgroundRunner : BackgroundRunner {
+    private var backgroundTask: UIBackgroundTaskIdentifier = .invalid
+    private var audioPlayer: AVAudioPlayer?
+
+    /// Invoke the given task in the background.
+    func invoke(background: @escaping ()  -> Void) {
+        let serialQueue = DispatchQueue(label: "org.openobservatory.queue", qos: .userInitiated, attributes: [.concurrent])
```
sdsantos, line 11: Shouldn't the label differ from depending on the applicationID or package? Or there's no possibility of conflict between apps?

sdsantos (approved):

iosApp/iosApp/background/IosBackgroundRunner.swift
aanorbel: ah. good point. Will correct this.

PR #268, opened 2024-11-13
fix: display proper test name in progress views

aanorbel:


composeApp/src/commonMain/kotlin/org/ooni/probe/data/disk/WriteFile.kt
```diff
@@ -29,17 +28,17 @@ class WriteFileOkio(
...
             absolutePath.parent?.let { fileSystem.createDirectories(it) }
-        } catch (e: IOException) {
+
+            fileSystem
+                .run { if (append) appendingSink(absolutePath) else sink(absolutePath) }
+                .use { sink ->
+                    sink.buffer().use {
+                        it.writeUtf8(contents)
+                    }
+                }
```
sdsantos, line 38: This is meant to fix the OutOfMemoryCrash? If so, likely the log message below needs to be different. Maybe we should split into 2 different try/catches. And the log message should probably be Logger.w or Logger.e so we get this on Sentry.

sdsantos (approved):

PR #271, opened 2024-11-13
update: explorer url

aanorbel:


composeApp/src/commonMain/release/kotlin/org/ooni/probe/config/OrganizationConfigInterface.kt
```diff
@@ -12,6 +12,9 @@ interface OrganizationConfigInterface {
     val ooniRunDashboardUrl: String
         get() = "https://run.ooni.org"
 
+    val explorerUrl: String
+        get() = "https://explorer.test.ooni.org"
```
sdsantos, line 16: The release shouldn't be the `test` one right?

sdsantos (changes requested):

composeApp/src/commonMain/release/kotlin/org/ooni/probe/config/OrganizationConfigInterface.kt
aanorbel: We want to keep it like this for release so we can see the changes being made to hide the header and footer in explorer

PR #272, opened 2024-11-13
Fix logging performance issues

### LINKED ISSUE #261: NMS Android: App crash and performance degrades
I will keep trying this to narrow in on exact steps to reproduce, but I was just able to make the app crash. Here's what I was up to immediate before the crash: 

- I went to settings>advanced
- Turned on debug logs
- Ran  the 'trusted media test'
- Put my phone down, phone went into sleep mode
- Picked up phone again, entered password to start using it 
- App screen was blank at first, took some time to load the UI of the app again
- After a few seconds it crashed
- After re-starting the app, it felt very laggy to use, there was a delay between tapping around on the nav and the UI updating

I was able to grab a screen recording so you can see some of this laggy-ness in action: 

https://photos.app.goo.gl/FQG6dnw3DVoPUQDv5

sdsantos:
Closes #261 

Basically delayed and throttled the actual file write.


aanorbel (approved):
🐳

Issue #174, opened 2024-10-11
update multiplatform app to use ooni descriptors.


aanorbel:
Note >> `composeApp/src/ooniMain/kotlin/org/ooni/probe/domain/GetBootstrapTestDescriptors.kt` and update `GetDefaultTestDescriptors`

jbonisteel:
Not for first release of multi-platform apps, we will discuss more and do this post launch

PR #275, opened 2024-11-18
feat: add ooni test descriptors.

aanorbel:
Closes https://github.com/ooni/probe-multiplatform/issues/173, https://github.com/ooni/probe-multiplatform/issues/174


composeApp/src/commonMain/kotlin/org/ooni/probe/data/models/InstalledTestDescriptorModel.kt
```diff
@@ -53,6 +58,21 @@ data class InstalledTestDescriptorModel(
 
     val isExpired get() = expirationDate != null && expirationDate < LocalDateTime.now()
 
+    val isDefaultTestDescriptor get() = id.value in 10470..10474 // TODO(aanorbel): switch to OONI reserved namespace
+
+    val key get() = if (isDefaultTestDescriptor) {
```
sdsantos, line 63: This is more like a settings key now no? Just to make it clear why we still need this. Because for all other purposes now we can use the ID.

sdsantos (changes requested):
I feel we should remove Defaults descriptors all together. If we don't have them anymore, we don't need the extra complexity. That means:
- No need to keep both `Descriptor` and `InstalledTestDescriptorModel` classes
- `Descriptor.Source` is not needed
- `DefaultTestDescriptor` is not needed
- `GetDefaultTestDescriptors` is not needed
- Only `isDefault()` method is needed, the `isInstalledNonDefaultDescriptor()` should just be `!isDefault()`.

That should simplify a lot of the existing code.

sdsantos:
> I feel we should remove Defaults descriptors all together. If we don't have them anymore, we don't need the extra complexity. That means:
> 
> * No need to keep both `Descriptor` and `InstalledTestDescriptorModel` classes
> * `Descriptor.Source` is not needed
> * `DefaultTestDescriptor` is not needed
> * `GetDefaultTestDescriptors` is not needed
> * Only `isDefault()` method is needed, the `isInstalledNonDefaultDescriptor()` should just be `!isDefault()`.
> 
> That should simplify a lot of the existing code.

Maybe we can keep both Descriptor and InstalledTestDescriptorModel. It may be useful for the `updateStatus` for example. But there's no need to duplicate fields, we can just have the installed model as a field of the descriptor.

composeApp/src/commonMain/kotlin/org/ooni/probe/data/models/RunSpecification.kt
```diff
@@ -17,24 +17,11 @@ sealed interface RunSpecification {
...
-
-            companion object {
-                fun fromDescriptor(descriptor: Descriptor) =
-                    when (descriptor.source) {
-                        is Descriptor.Source.Default -> Default(descriptor.name)
-                        is Descriptor.Source.Installed -> Installed(descriptor.source.value.id)
-                    }
-            }
+        companion object {
+            fun fromDescriptor(descriptor: Descriptor) = descriptor.source.id
```
sdsantos, line 24: This companion object isn't needed, it should return an object of the type `Test`, or it should be called something like `sourceFromDescriptor`. It's weird that a static factory method only returns a specific field inside its class.

composeApp/src/commonMain/kotlin/org/ooni/probe/data/models/RunSpecification.kt
```diff
@@ -17,24 +17,11 @@ sealed interface RunSpecification {
 
     @Serializable
     data class Test(
-        val source: Source,
+        val source: InstalledTestDescriptorModel.Id? = null,
```
sdsantos, line 20: Maybe call this sourceId or descriptorId to avoid confusion?

sdsantos: And this should never be null right?

composeApp/src/commonMain/kotlin/org/ooni/probe/ui/choosewebsites/ChooseWebsitesViewModel.kt
```diff
@@ -104,7 +104,6 @@ class ChooseWebsitesViewModel(
                     RunSpecification.Full(
                         tests = listOf(
                             RunSpecification.Test(
-                                source = RunSpecification.Test.Source.Default("websites"),
```
sdsantos, line 107: We should fetch it from the DB, or hardcode the websites descriptor ID somewhere.

composeApp/src/dwMain/kotlin/org/ooni/probe/config/OrganizationSettings.kt
sdsantos, line 1: Is this is empty, it can be removed no?

sdsantos (changes requested):
I found a bug. When you just have the OONI descriptors, and no custom descriptor installed, the pull-to-refresh on the Dashboard screen doesn't seem to work. The refresh icon just says at the top without animation.

composeApp/src/commonMain/kotlin/org/ooni/probe/data/models/InstalledTestDescriptorModel.kt
```diff
@@ -52,6 +72,10 @@ data class InstalledTestDescriptorModel(
 
     val isExpired get() = expirationDate != null && expirationDate < LocalDateTime.now()
 
+    val isDefaultTestDescriptor get() = id.value in 10470..10474 // TODO(aanorbel): switch to OONI reserved namespace
```
sdsantos, line 75: We could also check if they are in `OoniTest` now.

sdsantos (approved):

sdsantos (approved):

aanorbel:
This will be done eventually but converting to draft for now. https://openobservatory.slack.com/archives/GKGRFHXT7/p1733782481912289?thread_ts=1733329764.355889&cid=GKGRFHXT7

composeApp/src/commonMain/kotlin/org/ooni/probe/data/models/Descriptor.kt
```diff
@@ -22,31 +22,18 @@ data class Descriptor(
     val expirationDate: LocalDateTime?,
     val netTests: List<NetTest>,
     val longRunningTests: List<NetTest> = emptyList(),
-    val source: Source,
+    val source: InstalledTestDescriptorModel,
```
sdsantos, line 25: I had mentioned this before, but keeping both the `Descriptor` and the `InstalledTestDescriptorModel` is just a big duplication of fields now. If we are to keep both because it helps with `UpdateStatus`, then we should remove the duplication:
- The `Descriptor` class should just have the `source` and `UpdateStatus`
- Optionally, `InstalledTestDescriptorModel` could be renamed to `Descriptor` and `Descriptor` could become a `DescriptorItem` or `DescriptorWithUpdateStatus`

composeApp/src/commonMain/kotlin/org/ooni/probe/data/models/InstalledTestDescriptorModel.kt
```diff
@@ -24,6 +28,26 @@ import org.ooni.probe.shared.hexToColor
...
 
+enum class OoniTest(
+    val id: Long,
+    val key: String,
+) {
+    WEBSITES(10470L, "websites"),
+    INSTANT_MESSAGING(10471L, "instant_messaging"),
+    CIRCUMVENTION(10472L, "circumvention"),
+    PERFORMANCE(10473L, "performance"),
+    EXPERIMENTAL(10474L, "experimental"),
```
sdsantos, line 39: We're using regular capitalization for enum and sealed classes in this project: WEBSITES -> Websites.

composeApp/src/commonMain/kotlin/org/ooni/probe/data/models/InstalledTestDescriptorModel.kt
```diff
@@ -24,6 +28,26 @@ import org.ooni.probe.shared.hexToColor
...
+    ;
+
+    companion object {
+        private val map = entries.associateBy(OoniTest::id)
+
+        fun fromId(id: Long) = map[id]
+
+        fun isValidId(id: Long): Boolean = entries.any { it.id == id }
```
sdsantos, line 47: Since we're already caching the entries as a map, this could be `fromId(id) != null`.

composeApp/src/commonMain/kotlin/org/ooni/probe/data/models/InstalledTestDescriptorModel.kt
```diff
@@ -24,6 +28,26 @@ import org.ooni.probe.shared.hexToColor
 import org.ooni.probe.shared.stringMonthArrayResource
 import org.ooni.probe.shared.toEpoch
 
+enum class OoniTest(
```
sdsantos, line 31: This should be in a separate file, there's no need to be here.

composeApp/src/commonMain/kotlin/org/ooni/probe/data/models/InstalledTestDescriptorModel.kt
```diff
@@ -56,6 +80,8 @@ data class InstalledTestDescriptorModel(
         val revision: Long,
     )
 
+    val isDefaultTestDescriptor get() = OoniTest.isValidId(id.value.toLongOrNull() ?: -1L)
```
sdsantos, line 83: Not sure if we should keep with the default concept or not. It's confusing with the NewsMediaScan descriptors, which are default, but not OONI tests. Maybe split this between `isOoniDescriptor` and `isUninstallable`.

composeApp/src/commonMain/kotlin/org/ooni/probe/domain/descriptors/GetTestDescriptorsBySpec.kt
```diff
@@ -34,16 +34,13 @@ class GetTestDescriptorsBySpec(
     // Is this descriptor contained in the RunSpecification's list of tests
     private fun RunSpecification.Full.forDescriptor(descriptor: Descriptor) =
         tests.firstOrNull { specTest ->
-            when (descriptor.source) {
-                is Descriptor.Source.Default -> {
+            when (val source = descriptor.source) {
```
sdsantos, line 37: Not needed anymore.

sdsantos (changes requested):
We should also migrate the ResultModel to stop having `descriptorName` and `descriptorKey`, and just have `descriptorKey`. That will simplify the SQL queries we have that filter by descriptor.

composeApp/src/commonMain/kotlin/org/ooni/probe/domain/RunDescriptors.kt
```diff
@@ -150,23 +153,26 @@ class RunDescriptors(
...
+        descriptor: DescriptorItem,
         index: Int,
         taskOrigin: TaskOrigin,
         isRerun: Boolean,
         noInternetWatcher: NoInternetWatcher,
     ) {
         val result = ResultModel(
             descriptorName = descriptor.name,
-            descriptorKey = (descriptor.source as? Descriptor.Source.Installed)?.value?.key,
+            descriptorKey = Descriptor.Key(
```
sdsantos: Same as `descriptorKey = descriptorItem.descriptor.key`. I would argue that even `descriptorKey = descriptorItem.key` should work, but it seems it's not the same thing for DescriptorItem. We should make sure the `key` methods return the same between Descriptor and DescriptorItem.

composeApp/src/commonMain/kotlin/org/ooni/probe/domain/RunNetTest.kt
```diff
@@ -67,8 +67,8 @@ class RunNetTest(
                     testTotal = spec.testTotal,
                 )
             }
-            val installedDescriptorId =
-                (spec.descriptor.source as? Descriptor.Source.Installed)?.value?.id
+            // Ensure descriptor id is not available for ooni descriptors
+            val installedDescriptorId = if (spec.descriptor.isDefault()) null else spec.descriptor.descriptor.id
```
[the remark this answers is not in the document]
aanorbel: We have not been sending, it wont make sense to start adding now

PR #289, opened 2024-11-19
show website animation and colors in RunningScreen

aanorbel:
Closes https://github.com/ooni/probe-multiplatform/issues/274


composeApp/src/commonMain/kotlin/org/ooni/probe/domain/RunDescriptors.kt
```diff
@@ -71,19 +71,32 @@ class RunDescriptors(
...
             .filterNot { it.test is TestType.WebConnectivity && it.inputs?.any() != true }
 
-    private suspend fun NetTest.inputsOrDownloadUrls(taskOrigin: TaskOrigin): List<String>? {
+    private suspend fun NetTest.inputsOrDownloadUrls(
+        taskOrigin: TaskOrigin,
+        descriptor: Descriptor,
+    ): List<String>? {
         if (!inputs.isNullOrEmpty() || test !is TestType.WebConnectivity) return inputs
 
+        setRunBackgroundState {
```
sdsantos, line 93: This sort of assumes that the Websites descriptor will always be the first one to run. It's true for now, but we need to be careful.

sdsantos: We are looping through all descriptors and tests. There's no check if it's the first one I believe.

PR #303, opened 2024-11-25
Improve DescriptorUpdateWorker exception handling to avoid misleading reports

sdsantos:


aanorbel (approved):
🐳  pending format fix

PR #305, opened 2024-11-25
Test reporting improvements

sdsantos:
- Don't report ResolverLookupFailures
- Skip reporting StartupFailures if the user stopped/cancelled the test
- Report ResolverLookupFailure, StartupFailures and BugJsonDump always using an exception, to see if they are grouped together on Sentry
- Revert debug urls back to `dev`


aanorbel (approved):
🐳

PR #307, opened 2024-11-25
Fix first descriptor update on NewsMediaScan

sdsantos:
The main change is making sure `fetchDescriptorUpdate` is called after `bootstrapTestDescriptors` in the `App.kt`.


aanorbel (approved):
🐳

PR #315, opened 2024-11-27
Upload missing results on start

### LINKED ISSUE #248: Retry uploads on app start
> We should give a visual cue probably with the same progress bar we have for the running tests. We also should ideally not allow running tests while an upload is in progress to avoid running a test that interferes with the measurement upload.

sdsantos:
Closes #248


aanorbel (approved):
🐳

PR #328, opened 2024-12-03
Select/deselect all website categories

### LINKED ISSUE #323: OONI Probe iOS: No ability to select/deselect all for Website categories
Navigate to: 

- Settings
- Test Options
- Website categories to test

On old OONI Probe, in the upper-right hand corner there are three dots '...' when you are in the Website categories to test view, and upon tapping those, you see the options 'Select All, Deselect All, cancel' 

On the multi-platform OONI probe, we are missing the three dots and thus you cannot select/deselect all. Screenshots are of old probe and new probe respectively 

OLDooni
Newooni

sdsantos:
Closes #323


composeApp/src/commonMain/composeResources/values/strings-common.xml
```diff
@@ -237,6 +237,8 @@
     <string name="Common_Refresh">refresh</string>
     <string name="Common_Collapse">Collapse</string>
     <string name="Common_Expand">Expand</string>
+    <string name="Common_SelectAll">Select all</string>
+    <string name="Common_DeselectAll">Deselect all</string>
```
aanorbel, line 241:
```suggestion
  <string name="Settings_Websites_Categories_Selection_None">Deselect All</string>
  <string name="Settings_Websites_Categories_Selection_All">Select All</string>
```

aanorbel (approved):
🐳

PR #329, opened 2024-12-03
Disable websites descriptor if no categories are enabled

### LINKED ISSUE #324: OONI Probe iOS: Websites still being tested even when zero categories are enabled
How to reproduce (also see screen recording): 

- Navigate to test options
- Ensure that zero categories are enabled under 'Limit Websites test duration' 
- Navigate back to the dashboard
- Ensure the 'Websites' category is selected
- Run 1 test
- When you look at the results, there will still be a handful of websites tested, even though 0 categories have been enabled in the settings

Expected results:

- On the old OONI Probe, if you have 0 website categories selected, the website 'card' is greyed out on the dashboard and you cannot run the Websites test. I assume this is still the behaviour that we want, so perhaps we should 'grey out' the ability to select the websites category when there are 0 categories enabled. 

https://github.com/user-attachments/assets/91a4e32e-bbf0-467c-9170-4aa33bc2a85e

sdsantos:
Closes #324


aanorbel (approved):
🐳

PR #338, opened 2024-12-09
Suppress LookupHost failed error reports

sdsantos:
We were getting too many individual LookupHost failed reports on Sentry.


aanorbel (approved):
🐳

PR #352, opened 2024-12-12
feat: add counts to summary cards

aanorbel:
Closes https://github.com/ooni/probe-multiplatform/issues/341

### DIFF
--- a/composeApp/src/commonMain/composeResources/values/strings-common.xml
+++ b/composeApp/src/commonMain/composeResources/values/strings-common.xml
@@ -71,6 +71,10 @@
     <string name="TestResults_Summary_Performance_Hero_Upload">Upload</string>
     <string name="TestResults_Summary_Performance_Hero_Download">Download</string>
 
+    <string name="TestResults_Gbps">Gbit/s</string>
+    <string name="TestResults_Mbps">Mbit/s</string>
+    <string name="TestResults_Kbps">kbit/s</string>
+
     <string name="TestResults_Summary_Websites_Hero_Tested_Singular">Tested</string>
     <string name="TestResults_Summary_Websites_Hero_Tested_Plural">Tested</string>
 
--- a/composeApp/src/commonMain/composeResources/values/strings-common.xml
+++ b/composeApp/src/commonMain/composeResources/values/strings-common.xml
@@ -70,6 +70,12 @@
     <string name="TestResults_NotAvailable">N/A</string>
     <string name="TestResults_Summary_Performance_Hero_Upload">Upload</string>
     <string name="TestResults_Summary_Performance_Hero_Download">Download</string>
+
+
+    <string name="TestResults_Summary_Websites_Hero_Tested_Singular">Tested</string>
+    <string name="TestResults_Summary_Websites_Hero_Blocked_Singular">Blocked</string>
+    <string name="TestResults_Summary_Websites_Hero_Reachable_Singular">Accessible</string>
+
     <string name="Modal_ReRun_Title">Re-run test</string>
     <string name="Modal_ReRun_Websites_Title">You are about to re-test %1$s websites.</string>
     <string name="Modal_ReRun_Websites_Run">Run</string>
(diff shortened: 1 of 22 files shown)


composeApp/src/commonMain/sqldelight/org/ooni/probe/data/Measurement.sq
```diff
@@ -62,6 +62,14 @@ SELECT * FROM Measurement
 LEFT JOIN Url ON Measurement.url_id = Url.id
 WHERE Measurement.result_id = ?;
 
+countByResultId:
+SELECT
+    COUNT(*) AS total_measurements,
+    SUM(CASE WHEN is_rerun = 0 AND is_done = 1 AND is_failed = 0 AND is_anomaly = 0 THEN 1 ELSE 0 END) AS ok_measurements,
+    SUM(CASE WHEN is_rerun = 0 AND is_done = 1 AND is_failed = 1 THEN 1 ELSE 0 END) AS failed_measurements
+FROM Measurement
+WHERE result_id = ?;
```
sdsantos, line 71: Do we need the extra DB query? Can't we just get this values from the whole list?

composeApp/src/commonMain/kotlin/org/ooni/probe/data/models/ResultItem.kt
```diff
@@ -2,11 +2,14 @@ package org.ooni.probe.data.models
 
 import kotlin.time.Duration.Companion.seconds
 
+data class ResultCount(val total: Long, val failed: Long?, val succeeded: Long?)
```
sdsantos, line 5: Why could `failed` and `succeeded` be null?

sdsantos (approved):

composeApp/src/commonMain/kotlin/org/ooni/probe/data/models/ResultItem.kt
```diff
@@ -2,6 +2,10 @@ package org.ooni.probe.data.models
 
 import kotlin.time.Duration.Companion.seconds
 
+data class ResultCount(val total: Long, val succeeded: Long) {
+    val failed get() = total - succeeded
```
sdsantos, line 6: I just realised that we're mixing failed and anomalies here. 
Blocked = isAnomaly
Failed = isFailed

We shouldn't be counting failures inside the blocked count.

composeApp/src/commonMain/composeResources/values/strings-common.xml
```diff
@@ -58,6 +58,12 @@
     <string name="TestResults_NotAvailable">N/A</string>
     <string name="TestResults_Summary_Performance_Hero_Upload">Upload</string>
     <string name="TestResults_Summary_Performance_Hero_Download">Download</string>
+
+
+    <string name="TestResults_Summary_Websites_Hero_Tested_Singular">Tested</string>
+    <string name="TestResults_Summary_Websites_Hero_Blocked_Singular">Blocked</string>
+    <string name="TestResults_Summary_Websites_Hero_Reachable_Singular">Accessible</string>
```
sdsantos, line 65: There are different strings for circunvention, websites and instant messaging

sdsantos (changes requested):
I missed a couple of things on my first review. Left some comments. Also, the counts summary is not visible for Experimental, and for Performance it's a totally different thing (https://github.com/ooni/probe-multiplatform/issues/343).

composeApp/src/commonMain/composeResources/values/strings-common.xml
aanorbel: Its thesame string. I doubt we want to keep 12+ strings saying similar things.

aanorbel: https://github.com/ooni/probe-android/blob/1b8d017ff72633083697340d3e8e4b3f6820d498/app/src/main/java/org/openobservatory/ooniprobe/fragment/resultHeader/ResultHeaderTBAFragment.java


I will switch to using the plural

sdsantos (approved):

PR #353, opened 2024-12-13
feat: disable `ndt` and `dash` in background run

aanorbel:
Closes https://github.com/ooni/probe-multiplatform/issues/350

### DIFF
--- a/composeApp/src/commonMain/kotlin/org/ooni/probe/data/repositories/PreferenceRepository.kt
+++ b/composeApp/src/commonMain/kotlin/org/ooni/probe/data/repositories/PreferenceRepository.kt
@@ -147,7 +147,7 @@ class PreferenceRepository(
         isAutoRun: Boolean,
     ): Flow<Boolean> =
         dataStore.data.map {
-            it[booleanPreferencesKey(getNetTestKey(descriptor, netTest, isAutoRun))] == true
+            it[booleanPreferencesKey(getNetTestKey(descriptor, netTest, isAutoRun))] ?: netTest.defaultPreferenceValue(isAutoRun)
         }.distinctUntilChanged()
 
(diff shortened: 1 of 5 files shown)


composeApp/src/commonMain/kotlin/org/ooni/probe/data/repositories/PreferenceRepository.kt
```diff
@@ -147,7 +147,7 @@ class PreferenceRepository(
         isAutoRun: Boolean,
     ): Flow<Boolean> =
         dataStore.data.map {
-            it[booleanPreferencesKey(getNetTestKey(descriptor, netTest, isAutoRun))] == true
+            it[booleanPreferencesKey(getNetTestKey(descriptor, netTest, isAutoRun))] ?: netTest.defaultPreferenceValue(isAutoRun)
```
sdsantos, line 150: If we're already bootstrapping the preferences correctly, why do we the default here? That's a different approach from the other defaults we have. I feel we can skip this, and just have them set on `BootstrapPreferences`, but maybe I'm missing something.

sdsantos (commented):
The issue doesn't have description. But from the issue title, we don't want NDT and Dash to run during auto-run. But this PR isn't doing that, it's just making sure that they don't run during auto-run **by default**. Users can still enabled them for auto-run. Is that what we want?

composeApp/src/commonMain/kotlin/org/ooni/probe/data/repositories/PreferenceRepository.kt
aanorbel: We bootstrap references in multiplatform but the android app has default preferences that are not available on disk untill the user changes them.

sdsantos (approved):

Issue #356, opened 2024-12-18
Probe Android: Run v2 link is missing the expired tag in the app


jbonisteel:
https://run.ooni.org/v2/10107

This link is expired, but the expired 'tag' is not showing in the multiplatform app. See screenshot: 

I have tried refreshing the desktop and no change. 

![Screenshot_20241218-101748](https://github.com/user-attachments/assets/3c9ddc13-0752-487b-91ed-2a61f1059ecb)

PR #359, opened 2024-12-18
feat: add expired tag

aanorbel:
Closes https://github.com/ooni/probe-multiplatform/issues/356

### DIFF
--- a/composeApp/src/commonMain/kotlin/org/ooni/probe/domain/GetTestDescriptors.kt
+++ b/composeApp/src/commonMain/kotlin/org/ooni/probe/domain/GetTestDescriptors.kt
@@ -19,13 +19,13 @@ import org.ooni.probe.data.models.toDescriptor
 
 class GetTestDescriptors(
     private val getDefaultTestDescriptors: () -> List<DefaultTestDescriptor>,
-    private val listInstalledTestDescriptors: () -> Flow<List<InstalledTestDescriptorModel>>,
+    private val listInstalledTestDescriptors: (Boolean?) -> Flow<List<InstalledTestDescriptorModel>>,
     private val descriptorUpdates: () -> Flow<DescriptorUpdatesStatus>,
     private val getPreferenceValues: (List<SettingsKey>) -> Flow<Map<SettingsKey, Any?>>,
 ) {
-    operator fun invoke(): Flow<List<Descriptor>> {
+    operator fun invoke(isExpired: Boolean? = null): Flow<List<Descriptor>> {
         return combine(
-            listInstalledTestDescriptors(),
+            listInstalledTestDescriptors(isExpired),
             descriptorUpdates(),
             flowOf(getDefaultTestDescriptors()),
             isWebsitesDescriptorEnabled(),
--- a/composeApp/src/commonMain/kotlin/org/ooni/probe/domain/GetTestDescriptors.kt
+++ b/composeApp/src/commonMain/kotlin/org/ooni/probe/domain/GetTestDescriptors.kt
@@ -19,13 +19,13 @@ import org.ooni.probe.data.models.toDescriptor
 
 class GetTestDescriptors(
     private val getDefaultTestDescriptors: () -> List<DefaultTestDescriptor>,
-    private val listInstalledTestDescriptors: (Boolean?) -> Flow<List<InstalledTestDescriptorModel>>,
+    private val listInstalledTestDescriptors: () -> Flow<List<InstalledTestDescriptorModel>>,
     private val descriptorUpdates: () -> Flow<DescriptorUpdatesStatus>,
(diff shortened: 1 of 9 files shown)


composeApp/src/commonMain/kotlin/org/ooni/probe/domain/GetTestDescriptors.kt
```diff
@@ -19,13 +19,13 @@ import org.ooni.probe.data.models.toDescriptor
 
 class GetTestDescriptors(
     private val getDefaultTestDescriptors: () -> List<DefaultTestDescriptor>,
-    private val listInstalledTestDescriptors: () -> Flow<List<InstalledTestDescriptorModel>>,
+    private val listInstalledTestDescriptors: (Boolean?) -> Flow<List<InstalledTestDescriptorModel>>,
     private val descriptorUpdates: () -> Flow<DescriptorUpdatesStatus>,
     private val getPreferenceValues: (List<SettingsKey>) -> Flow<Map<SettingsKey, Any?>>,
 ) {
-    operator fun invoke(): Flow<List<Descriptor>> {
+    operator fun invoke(isExpired: Boolean? = null): Flow<List<Descriptor>> {
```
sdsantos, line 26: I don't think there's a case we'll want to get just the expired test descriptors with all the extra data this is fetching. So I think we could avoid having this argument here, so we don't need to set it as false in a lot of places.

composeApp/src/commonMain/sqldelight/org/ooni/probe/data/TestDescriptor.sq
```diff
@@ -71,7 +71,7 @@ setAutoUpdate:
 UPDATE TestDescriptor SET auto_update = ? WHERE runId = ?;
 
 selectAll:
-SELECT * FROM TestDescriptor;
+SELECT * FROM TestDescriptor WHERE (is_expired IS NULL OR is_expired = :isExpired) OR (:isExpired IS NULL);
```
sdsantos, line 74: Although we have a `is_expired` field inside the DB, we shouldn't. It's not a reliable way to know if the descriptor is actually expired, because it won't be updated every day. We only update it correctly if the user as enabled auto-update. Otherwise, it may be already expired, but the DB field is not updated.

Either we update the `is_expired` field for all descriptors consistently, or we should drop it from the DB and just use the expired date.

sdsantos (changes requested):

composeApp/src/commonMain/kotlin/org/ooni/probe/domain/GetTestDescriptors.kt
aanorbel: I think we need the argument. include expired tests or not to include expired tests

composeApp/src/commonMain/sqldelight/org/ooni/probe/data/TestDescriptor.sq
```diff
@@ -71,8 +71,7 @@ setAutoUpdate:
 UPDATE TestDescriptor SET auto_update = ? WHERE runId = ?;
 
 selectAll:
-SELECT * FROM TestDescriptor;
-
+SELECT * FROM TestDescriptor WHERE (is_expired IS NULL OR (is_expired = :isExpired AND expiration_date > strftime('%s', 'now'))) OR (:isExpired IS NULL);
```
sdsantos, line 74: I believe this query is wrong. It should be ignoring the `is_expired` DB column and it doesn't work for returning the expired descriptors. Maybe try writing some tests for it?

I was expecting something like:
```
SELECT * FROM TestDescriptor WHERE
:isExpired IS NULL 
OR (:isExpired = 1 AND expiration_date < strftime('%s', 'now'))
OR (:isExpired = 0 AND expiration_date >= strftime('%s', 'now'))
```

sdsantos (changes requested):

composeApp/src/commonMain/kotlin/org/ooni/probe/ui/run/RunViewModel.kt
```diff
@@ -278,3 +278,12 @@ class RunViewModel(
...
 }
+
+private fun List<Descriptor>.runnableDescriptors(): List<Descriptor> {
+    return filter {
+        when (it.source) {
+            is Descriptor.Source.Default -> true
+            is Descriptor.Source.Installed -> !it.isExpired
+        }
+    }
+}
```
sdsantos, line 289: The `Descriptor` model also has the `isExpired` field, so this `when` is not needed.

sdsantos (changes requested):
We also need to filter expired descriptors inside `GetAutoRunSpecification` at least.

sdsantos (approved):

PR #360, opened 2024-12-19
feat: Show performance metrics on performance items

aanorbel:
Closes https://github.com/ooni/probe-multiplatform/issues/342


composeApp/src/commonMain/kotlin/org/ooni/probe/data/repositories/MeasurementRepository.kt
```diff
@@ -54,6 +60,17 @@ class MeasurementRepository(
             .mapToList(backgroundContext)
             .map { list -> list.mapNotNull { it.toModel() } }
 
+    fun selectTestKeysByResultId(filter: ResultFilter): Flow<List<TestKeysWithResultId>> {
+        val descriptorFilter = (filter.descriptor as? ResultFilter.Type.One)?.value
+        return database.measurementQueries
+            .selectTestKeysByResultId(
```
sdsantos, line 66: It says byResultId by you are only filtering by descriptor. Are you already preparing for the other task showing the summary inside the result? Nevertheless, the name of the query feels wrong.

composeApp/src/commonMain/kotlin/org/ooni/probe/domain/RunNetTest.kt
```diff
@@ -163,6 +163,14 @@ class RunNetTest(
                             )
                         }
 
+                        event.result.testKeys?.let {
+                            val testKeys = extractTestKeysPropertiesToJson(it)
+                            Logger.d("testKeys: ${measurement.id}, $testKeys")
```
sdsantos, line 168: Is this log still needed, or a good idea to keep?

composeApp/src/commonMain/kotlin/org/ooni/probe/domain/EvaluateMeasurementKeys.kt
```diff
@@ -92,3 +92,17 @@ fun evaluateMeasurementKeys(
...
+            "median_bitrate" to testKeys.simple?.medianBitrate,
+            "upload" to testKeys.simple?.medianBitrate,
+            "download" to testKeys.simple?.medianBitrate,
+        ),
+        "summary" to mapOf(
+            "upload" to testKeys.summary?.upload,
+            "download" to testKeys.summary?.download,
+        ),
+    )
+}
```
sdsantos, line 108: We could avoid storing the null values, to save on DB space.

sdsantos (changes requested):
We should avoid showing the icons when there's no info:

Screenshot 2025-01-06 at 11 54 31

The icons and text should be smaller like the other info we're showing:

Screenshot 2025-01-06 at 12 41 39

composeApp/src/commonMain/kotlin/org/ooni/probe/domain/EvaluateMeasurementKeys.kt
```diff
@@ -92,3 +92,21 @@ fun evaluateMeasurementKeys(
                     keys?.registrationServerStatus == TestKeys.BLOCKED_VALUE,
             )
     }
+
+fun extractTestKeysPropertiesToJson(testKeys: TestKeys): Map<String, Map<String, Double?>?> {
+    return mapOf(
```
sdsantos, line 97: This `mapOf` could also use `.filter { it.value != null }`

sdsantos (approved):

PR #371, opened 2025-01-08
Background work documentation

sdsantos:
I also noticed that our Android workers didn't have the `NetworkType.CONNECTED` and I believe they should have. There's nothing to test or update without any kind of Internet connection.


aanorbel (approved):
🐳

PR #374, opened 2025-01-09
feat: Show performance metrics (video, download, upload and ping) on summary card 

aanorbel:
Closes https://github.com/ooni/probe-multiplatform/issues/343


sdsantos (approved):
I would just recommend a single decimal point and slightly smaller text font for the values, to make sure they fit horizontally on the screen.

composeApp/src/commonMain/kotlin/org/ooni/probe/domain/GetResults.kt
```diff
@@ -12,13 +12,13 @@ import org.ooni.probe.data.models.TestKeysWithResultId
 class GetResults(
     private val getResults: (ResultFilter) -> Flow<List<ResultWithNetworkAndAggregates>>,
     private val getDescriptors: () -> Flow<List<Descriptor>>,
-    private val getTestKeys: (ResultFilter) -> Flow<List<TestKeysWithResultId>>,
+    private val getTestKeys: (String?) -> Flow<List<TestKeysWithResultId>>,
```
sdsantos, line 15: Not a bit thing, but next time use the `ResultModel.Id` as the argument, cause String can be anything.

sdsantos (approved):

composeApp/src/commonMain/kotlin/org/ooni/probe/domain/GetResults.kt
aanorbel: I am going to merge as is. It felt a bit weired passing `ResultFilter` as an argument and also instantiating `ResultModel.Id`. But i agree. Better to instantiate `ResultModel.Id`

PR #402, opened 2025-01-20
Improve StartupFailure/ResolverLookupFailure message

### LINKED ISSUE #395: Disaggregate MKExceptions reported on Sentry based on their failure message
Currently they are being reported as `RunNetTest.StartupFailure` and `ResolverLookupFailure` and the failure message is already being taken in account. But we're joining the json value, and that does not seem to be needed.

sdsantos:
Closes #395

We are already splitting the MKException into StartupFailure/ResolverLookupFailure, and we're taking the failure message from the event value. Here's how the errors currently look on Sentry:
https://ooni.sentry.io/issues/6216424916/?project=4508325650235392
https://ooni.sentry.io/issues/6225238701/?project=4508325650235392

I just improve the messaging, because we don't need the json if we have the failure message.

### DIFF
--- a/composeApp/src/commonMain/kotlin/org/ooni/probe/domain/RunNetTest.kt
+++ b/composeApp/src/commonMain/kotlin/org/ooni/probe/domain/RunNetTest.kt
@@ -304,15 +304,8 @@ class RunNetTest(
         )
     }
 
-    open inner class Failure(message: String?, value: TaskEventResult.Value?) : Exception(
-        if (message != null && value != null) {
-            message + "\n" + json.encodeToString(value)
-        } else if (value != null) {
-            json.encodeToString(value)
-        } else {
-            message ?: ""
-        },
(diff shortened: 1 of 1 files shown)


aanorbel (commented):
I still think they would be grouped under `org.ooni.probe.domain.RunNetTest.StartupFailure`

We could create a sealed class for each possible startup failure so we have something in the form

- org.ooni.probe.domain.RunNetTest.StartupFailure.torsf
- org.ooni.probe.domain.RunNetTest.StartupFailure.vanila_tor
- org.ooni.probe.domain.RunNetTest.StartupFailure....

This would mean they can never be grouped. As a fallback, we can use the message.

> https://docs.sentry.io/product/issues/grouping-and-fingerprints/#fingerprint-rules

aanorbel (approved):
🐳

PR #408, opened 2025-01-21
Foreground Service Mitigations

### LINKED ISSUE #405: ForegroundService issue mitigation
- Make sure we’re calling `setForeground` as early as possible inside the `RunWorker` and the `DescriptorUpdateWorker`
- Specify the foreground service type `dataSync` in the `ForegroundInfo`
- Handle `onTimeout(int, int)`, call `stopSelf()` inside and log the timeout to Sentry

sdsantos:
Closes #405 

setForeground on workers with foreground service type.
Handle restrictions and early stops. Timeouts and other early stops are catched as CancellationException. 
Also tried to make logging more similar across the 2 workers.


aanorbel (approved):
🐳

PR #411, opened 2025-01-22
More detailed uninstall link prompt string

### LINKED ISSUE #384: Run V2 - Update uninstall link confirmation dialog text, to include a mention about also deleting the results

sdsantos:
Closes #384


aanorbel (approved):
Looks like too many modified translations. 

Was that intended?

sdsantos:
> Looks like too many modified translations.
> 
> Was that intended?

I modified all translations that included that string, I thought that was the plan? The rest of the changes were automatic whitespace changes.

aanorbel:
We can go ahead and merge. the other language files are generated so i think the content would be replaced when an update is made from transifex

PR #424, opened 2025-01-28
chore(iOS): restructure background task

### LINKED ISSUE #399: [iOS] Report to Sentry when our background services are killed by the operative system
We should try to include as much data as we can in the report, like how long was the background service running for.

aanorbel:
Closes #399

### DIFF
--- a/composeApp/src/iosMain/kotlin/org/ooni/probe/background/OperationsManager.kt
+++ b/composeApp/src/iosMain/kotlin/org/ooni/probe/background/OperationsManager.kt
@@ -18,7 +18,7 @@ class OperationsManager(private val dependencies: Dependencies, private val back
     // https://developer.apple.com/documentation/foundation/nsdate/1410206-timeintervalsince
     fun startSingleRun(spec: RunSpecification) {
         val startDateTime = NSDate()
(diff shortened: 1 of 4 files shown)


composeApp/src/iosMain/kotlin/org/ooni/probe/background/OperationsManager.kt
```diff
@@ -1,26 +1,57 @@
...
+            } catch (e: Exception) {
+                Logger.e(e) { "Job failed with exception in ${NSDate().timeIntervalSinceDate(startDateTime)} ms" }
+            }
+        }
         backgroundRunner(background = {
-            runBlocking { dependencies.runBackgroundTask(spec).collect() }
+            differed.start()
+        }, cancel = {
+            differed.cancel()
+            Logger.i { "Job cancelled after ${NSDate().timeIntervalSinceDate(startDateTime)} ms" }
```
sdsantos, line 33: `Logger.i` will only log as a breadcrumb. This only gets sent if there's an error. Or will we always call line 26 inside the `catch` when it's aborted?

aanorbel: Nope only on error.

PR #425, opened 2025-01-28
Persist rejected descriptor updates

### LINKED ISSUE #385: Run V2 - Rejected updates should be persisted
But there should be a way to review again rejected updates inside the Descriptor screen.

Tasks:
- [x] Add a explicit reject button in the update descriptor review screen
- [x] Store rejected "revision" numbers in the database or preferences, instead of in memory
- [x] In the descriptor screen, show last rejected revision number, with a button to review and install

sdsantos:
Closes #385 

Made sure descriptor update rejections are explicit, and stored them in the database. Users can undo the rejection in the descriptor screen.

Also refactored the code around the FetchDescriptorUpdates, to limit each class responsibilities.

|.|.|
|-|-|
|Screenshot_20250128_173737|Screenshot_20250128_173756|


aanorbel (approved):
🐳

Issue #435, opened 2025-02-04
probe-multiplatform: links to echcheck test measurements on explorer miss the input field


hellais:
In order to access echcheck measurement the URL needs to be specified as `report_id` and `input`, while only the `report_id` is getting passed.

Here is an example of how the URL should look like:
https://explorer.ooni.org/measurement/20250204T085731Z_echcheck_IT_30722_n1_TqovxoWAH3W7ViWo?input=https://cloudflare-ech.com/cdn-cgi/trace

While in the current app it's specified as:

https://explorer.ooni.org/measurement/20250204T085731Z_echcheck_IT_30722_n1_TqovxoWAH3W7ViWo

PR #437, opened 2025-02-04
feat: add support for url in `echcheck`

aanorbel:
Closes https://github.com/ooni/probe-multiplatform/issues/435


sdsantos (approved):
It would have been nice a comment explaining why we also need this here.

aanorbel:
> It would have been nice a comment explaining why we also need this here.

Adding a comment now.

PR #439, opened 2025-02-04
Don't fetch previous revisions from API

### LINKED ISSUE #438: Don't fetch previous revisions from API
Instead of fetching the list of previous revisions from the API, we can assume that revisions are always incremental, starting from 1, and never skip any number. This should simplify our logic, reduce the time to check for descriptor updates, and reduce the amount of API calls made.

sdsantos:
Closes #438


aanorbel (approved):
🐳

PR #442, opened 2025-02-04
Add F-Droid and App Store release badges

sdsantos:
Help us keep track of what versions of the app were released on F-Droid and Apple App Store.

Source: https://shields.io/badges


aanorbel (approved):
🐳

PR #444, opened 2025-02-05
Accessibility improvements

sdsantos:
Fix the issues I found using the Accessibility Scanner for our main screens:
- Lack of contrast between our primary and primaryContainer colors and the background. Contrast is now `4.54:1` between #0579B3 and #F7F9FF.
- Duplicated content description for the collapse/expand buttons on the Run screen
- Touch targets too small for the collapse/expand buttons and the test items on the Run screen


aanorbel (approved):
🐳

PR #457, opened 2025-02-11
Handle measurement webview errors

### LINKED ISSUE #454: probe-multiplatform: properly handle lack of connectivity when loading measurement pages
When I load a measurement page I get the stock android web view error message. We should instead see to handle this better and display a contextual error message stating that network connectivity is required in order to view results.

sdsantos:
Closes #454 

Screenshot 2025-02-11 at 14 00 24


aanorbel (approved):
🐳

Issue #443, opened 2025-02-05
Add metadata folder for F-Droid release


sdsantos:
F-Droid should pick up the app details from a `metadata` folder inside the repository. That's the only way of adding an app icon.

Reference: https://f-droid.org/en/docs/All_About_Descriptions_Graphics_and_Screenshots/

PR #470, opened 2025-02-12
chore: add metadata folder for F-Droid release

aanorbel:
Closes https://github.com/ooni/probe-multiplatform/issues/443


fastlane/.gitignore
```diff
@@ -2,6 +2,14 @@
 /report.xml
 /screenshots/*/*.png
 /screenshots/screenshots.html
```
sdsantos, line 4: Do we still need these?

sdsantos (approved):

fastlane/.gitignore
aanorbel: Yes. We dont want to track `screenshots.html` which is not excluded in any other ignore line.

aanorbel: I dont think so. Its just legacy ingore copied from fastlane

PR #473, opened 2025-02-12
chore: update fastlane config for `nms`

aanorbel:


fastlane/.gitignore
```diff
@@ -3,5 +3,6 @@
 /screenshots/*/*.png
 /screenshots/screenshots.html
 /screenshots/report.xml
+/metadata/dw/
```
sdsantos, line 6: This is already covered by the ´/metadata/` line below

sdsantos (approved):

fastlane/.gitignore
aanorbel: yeah. that line is removed in a different PR.

PR #488, opened 2025-02-17
Re-create DB if older than V5

sdsantos:
I copied the `CREATE TABLE` statements from the `5.db` file we have a reference. Then `4.sqm` migration runs when migrating from v4 to v5.


aanorbel (approved):
🐳  LGTM

PR #489, opened 2025-02-17
Integrate fastlane Google Play strings with transifex

sdsantos:


sdsantos:
@aanorbel Added the iOS base info for OONI and NMS too.

PR #490, opened 2025-02-17
Remove DependencyInfoBlock

licaon-kter:
```
$ apksigtool parse --json OONIProbe-Android-5.0.2-fdroid.apk | jq -r '.pairs[].value._type'
APKSignatureSchemeBlock
DependencyInfoBlock
VerityPaddingBlock
```

We find that there is a `DependencyInfoBlock` in your APK.

It's a [Signing block](https://source.android.com/docs/security/features/apksigning/v2#apk-signing-block) added by AGP and encrypted with the Google public key so it can't be read by anyone else except Google. You can read more about it [here](https://gitlab.com/fdroid/admin/-/issues/367) and [here](https://gitlab.com/fdroid/fdroidserver/-/issues/1056).

While OONI Probe is not built reproducible in F-Droid, we rebuild apps and try to fix issues: https://verification.f-droid.org so more apps are reproducible in general.

related: https://f-droid.org/docs/Inclusion_How-To/#reproducible-builds

related: https://gitlab.com/fdroid/fdroiddata/-/merge_requests/19716

### DIFF
--- a/composeApp/build.gradle.kts
+++ b/composeApp/build.gradle.kts
@@ -219,8 +219,8 @@ android {
     dependenciesInfo {
         // Disables dependency metadata when building APKs.
         includeInApk = false
-        // Disables dependency metadata when building Android App Bundles.
-        includeInBundle = false
+        // Keeps dependency metadata when building Android App Bundles.
+        includeInBundle = true
     }
     testOptions {
         execution = "ANDROIDX_TEST_ORCHESTRATOR"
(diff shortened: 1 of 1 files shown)


sdsantos:
@licaon-kter F-droid doesn't use the Android App Bundle right? Maybe we could leave it enabled for AABs and disable just for APKs?

licaon-kter:
@sdsantos sure :+1:

PR #511, opened 2025-02-19
CD: Workflow to publish android apps on google play

sdsantos:
TODO:

- [x] Generate Google Play service accounts for OONI and add it to Github Secrets
- [ ] Generate Google Play service accounts for NMS and add it to Github Secrets
- [x] Modify `composeApp/build.gradle` to allow external signing configs (instead of just using debug)
- [x] Add OONI signing configs 
- [ ] Add NMS signing configs 
- [ ] Test


aanorbel (approved):
🐳

PR #520, opened 2025-02-26
Chunk large URL IN SQL queries, to avoid crash

### LINKED ISSUE #519: Investigate UrlRepository SQLiteException: too many SQL variables
Crash report https://ooni.sentry.io/issues/6333137392/?project=4508325642764288&query=%21is%3Aarchived%20is%3Aunresolved%20installerStore%3A%5Bcom.huawei.appmarket%2Ccom.android.vending%5D&referrer=issue-stream&sort=freq&statsPeriod=7d&stream_index=7

sdsantos:
Closes #519

### DIFF
--- a/composeApp/src/commonMain/kotlin/org/ooni/probe/data/repositories/UrlRepository.kt
+++ b/composeApp/src/commonMain/kotlin/org/ooni/probe/data/repositories/UrlRepository.kt
@@ -42,7 +42,7 @@ class UrlRepository(
                 val urlsWithoutId = models.filter { it.id == null }.map { it.url }
                 val existingModels: List<UrlModel> =
                     // Some lists are too large for a single SQL query
-                    urlsWithoutId.chunked(300) { urlsChunk ->
+                    urlsWithoutId.chunked(200) { urlsChunk ->
                         database.urlQueries.selectByUrls(urlsChunk).executeAsList()
                     }
                         .flatMap { list -> list.mapNotNull { it.toModel() } }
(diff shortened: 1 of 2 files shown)


composeApp/src/commonMain/kotlin/org/ooni/probe/data/repositories/UrlRepository.kt
```diff
@@ -39,10 +39,13 @@ class UrlRepository(
...
         withContext(backgroundContext) {
             database.transactionWithResult {
-                val existingModels =
-                    database.urlQueries.selectByUrls(models.filter { it.id == null }.map { it.url })
-                        .executeAsList()
-                        .mapNotNull { it.toModel() }
+                val urlsWithoutId = models.filter { it.id == null }.map { it.url }
+                val existingModels: List<UrlModel> =
+                    // Some lists are too large for a single SQL query
+                    urlsWithoutId.chunked(300) { urlsChunk ->
```
aanorbel, line 45: where are we getting this limit from

aanorbel (approved):
🐳

composeApp/src/commonMain/kotlin/org/ooni/probe/data/repositories/UrlRepository.kt
sdsantos: It's random. I only crashed with 3000 URLs, so I assumed 300 was ok for SQL. We know we're fine with the 222 from NMS. Do you think we should chunk by another value?

And it's not a limit, we're just batching the queries in chunks.

aanorbel: The idea is to have a value that is proven to work.

PR #521, opened 2025-02-26
Document new release process. Create Github actions for it.

sdsantos:
My attempt at documenting the new Release process, and create the missing Github Actions to support it. And now that the Fastlane actions all have good descriptions, I think we can commit the fastlane README.


aanorbel (approved):
🐳

PR #536, opened 2025-02-27
Fix fastlane locales with symlinks

sdsantos:
There was also an issue with generating screenshots, because the package name was missing .dev.


aanorbel (approved):
🐳

PR #547, opened 2025-03-03
Fix metadata locales

sdsantos:
We were having issues with locales from Transifex and Google Play or the Apple App Store:
- fastlane didn't support having unknown locales in its metadata folder
- some locales use different names for each store: `es` or `es-ES` for example
- we also had situations like `pt-rBR` needed to become `pt-BR` (transifex doesn't let us have different locale mappings for different files)

So, to address this, I created a new root `metadata` folder that will sync with Transifex. And inside the `fastlane/metadata` folder we only have symbolic links to the locales we want in the stores. It works both ways. Screenshots taken by fastlane end up in the root `metadata` folder too.

Not sure if this solves the default changelogs upload. If not, I'll need to create symbolic links for the default changelogs file for each locale 😢.


aanorbel:
I understand the problem being solved here. 
We need to account for `f-droid` metadata and flavor placement. https://gitlab.com/fdroid/fdroidserver/-/issues/829

Will go through and try to make some suggestions.

aanorbel:
Were going to take the `fdroidclient` repo as a template. https://gitlab.com/fdroid/fdroidclient/-/tree/master/src . They have 2 flavors `full` and `basic`. 

We would need to emulate 2 flavors `ooni` and `dw` that serves the same function.

- `src/ooni/fastlane/metadata/android`
- `src/dw/fastlane/metadata/android`

The path above is required so f-droid can pick it up. Refer to https://gitlab.com/fdroid/fdroidserver/-/issues/829 for more information

We can add something like this to allow f-droid builld by the flavor name or find a way to include the `gradle` property as a flag for f-droid builds.

```kotlin

tasks.register("assembleOoniRelease", Exec::class) {
    group = "build"
    description = "Assembles a release build with OONI branding"
    workingDir = rootDir

    val isWindows = System.getProperty("os.name").lowercase().contains("windows")
    val gradlewExecutable = if (isWindows) "gradlew.bat" else "./gradlew"

    commandLine(
        gradlewExecutable,
        "assembleFdroidRelease",
        "-Porganization=ooni"
    )
}

tasks.register("assembleDwRelease", Exec::class) {
    group = "build"
    description = "Assembles a release build with DW branding"
    workingDir = rootDir

    val isWindows = System.getProperty("os.name").lowercase().contains("windows")
    val gradlewExecutable = if (isWindows) "gradlew.bat" else "./gradlew"

    commandLine(
        gradlewExecutable,
        "assembleFdroidRelease",
        "-Porganization=dw"
    )
}

```

sdsantos:
Added a symbolic link to make `fastlane/metadata/android` point to `fastlane/metadata/ooni/android` for f-droid.

aanorbel (approved):
LGTM 🐳

PR #553, opened 2025-03-04
Fix messaging test keys evaluation

sdsantos:
We were using the wrong keys for Telegram. And on Explorer I'm seeing lowercase "blocked", so I'm making sure our string checks ignore case.


aanorbel (approved):
🐳

PR #559, opened 2025-03-06
fix: app gallery deploy

aanorbel:
Closes https://github.com/ooni/probe-multiplatform/issues/535


sdsantos (approved):
Not sure if you tested with this locale format, but here they are using underscores instead of hyphens: https://developer.huawei.com/consumer/en/doc/app/agc-help-supported-countries-overview-0000001146718725

aanorbel:
the problematic languages were corrected.

PR #570, opened 2025-03-10
chore: generate ios screenshots.

aanorbel:
Closes https://github.com/ooni/probe-multiplatform/issues/534

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -243,7 +243,7 @@ platform :ios do
             scheme: "#{scheme}UITests",
             test_target_name: scheme,
             erase_simulator: true,
-            # languages: locales,
+            languages: locales,
             override_status_bar_arguments: "--time 12:00 --batteryState charging",
         )
     end
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -240,7 +240,8 @@ platform :ios do
         capture_screenshots(
             workspace: "iosApp/iosApp.xcworkspace",
             output_directory: "metadata/#{organization}/ios",
-            scheme: scheme,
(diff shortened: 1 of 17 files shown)


fastlane/Fastfile
```diff
@@ -233,7 +233,19 @@ platform :ios do
...
+        locales = locales_or_default(options)
+        scheme = scheme!(organization)
+
+        capture_screenshots(
+            workspace: "iosApp/iosApp.xcworkspace",
+            output_directory: "metadata/#{organization}/ios",
+            scheme: "#{scheme}UITests",
+            test_target_name: scheme,
+            erase_simulator: true,
+            # languages: locales,
```
sdsantos, line 246: Don't we need this?

metadata/.gitignore
```diff
@@ -12,3 +12,9 @@ dw/android/*/images/phoneScreenshots/*.png
 !dw/android/*/images/phoneScreenshots/3_*
 !dw/android/*/images/phoneScreenshots/4_*
 !dw/android/*/images/phoneScreenshots/5_*
+
+ooni/ios/*/*.png
+ooni/ios/*.html
+
+dw/ios/*/*.png
+dw/ios/*.html
```
sdsantos, line 20: Don't we want to eventually upload the screenshots into git?

sdsantos (approved):
I'm guessing this is a first pass, since we don't have the screenshots code working yet, right?

metadata/.gitignore
aanorbel: We do. But these are general rules. we generate more screenshots that we actually need to upload to git.

PR #571, opened 2025-03-10
Fix Android isBatteryCharging and auto-run notifications

sdsantos:
The `RunBackgroundTask.kt` changes fix the auto-run notification not updating.

I updated the `checkBatteryCharging` code to the code provided by the reference. But please test it too @aanorbel.

I also improved the descriptor update notification and notification channel. That introduced a new string for translations.

---

@agrabeli Can you approve the new "Link Updates" string we added? It's only visible through app settings. It's to distinguish notifications from our background tests, and when we're just updating the Links (once per day). Users can pick whether to see or not those notifications. The channel name needs to be short.


aanorbel (approved):
Tested.  🐳

PR #582, opened 2025-03-12
Fix IllegalArgumentException with duplicated URLs on RunScreen

sdsantos:
Crash report: https://ooni.sentry.io/issues/6386937398/?project=4508325642764288&query=%21is%3Aarchived%20is%3Aunresolved%20installerStore%3A%5Bcom.huawei.appmarket%2Ccom.android.vending%5D&referrer=issue-stream&sort=freq&statsPeriod=7d&stream_index=17

The app crashes on the NMS Run screen if the user expands the dropdown of the second or third option, before the app could update the descriptors. Removed the duplicated URLs because they made no sense, and protected against the crash if it happens again.

I want to include this in the current ongoing release.


aanorbel (approved):
🐋

PR #593, opened 2025-03-19
Fix check-in run type

sdsantos:


composeApp/src/commonMain/kotlin/org/ooni/engine/Engine.kt
```diff
@@ -96,7 +96,10 @@ class Engine(
                         charging = isBatteryCharging(),
                         onWiFi = networkTypeFinder() == NetworkType.Wifi,
                         platform = platformInfo.platform.value,
-                        runType = taskOrigin.value,
+                        runType = when (taskOrigin) {
+                            TaskOrigin.AutoRun -> "timed"
+                            TaskOrigin.OoniRun -> "manual"
+                        },
```
aanorbel, line 102: We should probably make this some type of property of the `TaskOrigin` >> `taskOrigin.runType` mabe.

aanorbel (approved):

composeApp/src/commonMain/kotlin/org/ooni/engine/Engine.kt
aanorbel:
```kotlin
fun TaskOrigin.runType(): String {
    return when (this) {
        TaskOrigin.AutoRun -> "timed"
        TaskOrigin.OoniRun -> "manual"
    }
}
```

PR #602, opened 2025-03-24
update iOS screenshots

aanorbel:


iosApp/iosApp/iOSApp.swift
```diff
@@ -16,6 +16,11 @@ class AppDelegate: NSObject, UIApplicationDelegate {
                                           showAlertAfterCurrentVersionHasBeenReleasedForDays: 0)
         siren.wail()
 
+        let launchArguments = ProcessInfo.processInfo.arguments
+
+        if launchArguments.contains("--skipOnboarding") {
+            UserDefaults.standard.set(3, forKey: "notification_popup")
+        }
```
sdsantos, line 23: I think this should have a comment explaining what is going on.

composeApp/src/iosMain/kotlin/org/ooni/probe/SetupDependencies.kt
```diff
@@ -64,6 +73,56 @@ import platform.UIKit.popoverPresentationController
 import platform.darwin.NSObject
 import platform.darwin.NSObjectMeta
 
+object PreferenceMigration : DataMigration<Preferences> {
```
sdsantos, line 76: This is applied not just for testing right? I feel like this will override our `BootstrapPreferences`, because it sets `FIRST_RUN`. Can't we just run this for testing? Through some flag like you have?

iosApp/iosApp/iOSApp.swift
aanorbel: Adding a description

composeApp/src/iosMain/kotlin/org/ooni/probe/SetupDependencies.kt
aanorbel: This is not for testing. To ensure that i can bypass onboarding, I needed to find a way to migrate native preferences to multiplatform which we did not do before. One way forward was to roll a preference migration by hand. Applies only to apps that dont have multiplatform preferences already.

sdsantos: Ok, you confirmed the migration doesn't run for new users? Then it's ok.

sdsantos (approved):

PR #621, opened 2025-03-25
Desktop battery state support

### LINKED ISSUE #608: desktop: isBatteryCharging
Figure out if there's a way for us to tell if a laptop is charging, or if we should default to `true` or a new `N/A` state.

@hellais mentioned it should be always true for desktop.

sdsantos:
Closes #608

### DIFF
--- a/composeApp/src/commonMain/kotlin/org/ooni/probe/data/models/BatteryState.kt
+++ b/composeApp/src/commonMain/kotlin/org/ooni/probe/data/models/BatteryState.kt
@@ -1,5 +1,7 @@
 package org.ooni.probe.data.models
 
-data class BatteryState(
-    val isCharging: Boolean,
-)
+enum class BatteryState {
+    NotCharging,
+    Charging,
+    Unknown,
+}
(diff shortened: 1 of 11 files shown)


composeApp/src/commonMain/kotlin/org/ooni/probe/data/models/BatteryState.kt
```diff
@@ -1,5 +1,7 @@
 package org.ooni.probe.data.models
 
-data class BatteryState(
-    val isCharging: Boolean,
-)
+enum class BatteryState {
+    NotCharging,
+    Charging,
```
aanorbel, line 5: Maybe its better to check represent this as plugged into AC power. It happens that the device is plugged in but not charging.

aanorbel (approved):
🐳

composeApp/src/commonMain/kotlin/org/ooni/probe/data/models/BatteryState.kt
sdsantos: But the app could be running on a laptop that may or not be charging. And it's not easy to know that for us. That's why I want with Unknown.

aanorbel: I see your point

PR #622, opened 2025-03-25
Desktop file directories

### LINKED ISSUE #604: desktop: set files / cache folders for each platform

sdsantos:
Closes #604 

Used https://codeberg.org/dirs/directories-jvm

### DIFF
--- a/composeApp/src/desktopMain/kotlin/org/ooni/probe/BuildDependencies.kt
+++ b/composeApp/src/desktopMain/kotlin/org/ooni/probe/BuildDependencies.kt
@@ -21,7 +21,7 @@ import org.ooni.probe.di.Dependencies
 import org.ooni.probe.shared.Platform
 import org.ooni.probe.shared.PlatformInfo
 
-private val projectDirectories = ProjectDirectories.from(null, null, "OONI Probe")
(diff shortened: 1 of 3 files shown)


composeApp/src/desktopMain/kotlin/org/ooni/probe/BuildDependencies.kt
```diff
@@ -20,14 +21,13 @@ import org.ooni.probe.di.Dependencies
 import org.ooni.probe.shared.Platform
 import org.ooni.probe.shared.PlatformInfo
 
-// TODO: Desktop - proper directories for each system
-private const val BASE_FILES_DIR = "../temp"
+private val projectDirectories = ProjectDirectories.from(null, null, "OONI Probe")
```
aanorbel, line 24: I think we should be ussing the package name here.

aanorbel (approved):
🐳

composeApp/src/desktopMain/kotlin/org/ooni/probe/BuildDependencies.kt
sdsantos: I'll go with the full set of arguments then:

Screenshot 2025-03-31 at 11 03 38

PR #631, opened 2025-03-31
ProxyScreen: add BackHandler

Lanius-collaris:
In v5.0.3, `ProxyViewModel.Event.BackClicked` isn't fired when users press the back button, the only way to save the custom proxy is clicking the ArrowBack icon (?) .

Fix ooni/probe#2850

Note: I'm not familiar with Compose.


sdsantos:
switched to https://github.com/ooni/probe-multiplatform/pull/635

PR #638, opened 2025-04-01
feat: add support for `ooni://` deeplinks on desktop

aanorbel:
Related with https://github.com/ooni/probe-multiplatform/issues/616

This is for macOS only. Need to integrate [unique4j](https://github.com/prat-man/unique4j) for [ complete implementation](https://conveyor.hydraulic.dev/17.0/configs/os-integration/#url-handlers-deep-linking)

NOTE: The app has to be installed to be tested.

use `open ooni://runv2/10004` from terminal or open link in the browser

### DIFF
--- a/composeApp/src/commonMain/resources/assets/.gitignore
+++ b/composeApp/src/commonMain/resources/assets/.gitignore
@@ -1,2 +1,0 @@
-
-descriptors.json
--- a/composeApp/src/commonMain/resources/assets/.gitignore
+++ b/composeApp/src/commonMain/resources/assets/.gitignore
@@ -1,0 +1,2 @@
+
+descriptors.json
--- a/composeApp/src/desktopMain/kotlin/org/ooni/probe/Main.kt
+++ b/composeApp/src/desktopMain/kotlin/org/ooni/probe/Main.kt
@@ -25,16 +25,19 @@ import org.ooni.probe.data.models.DeepLink
 import java.awt.Desktop
 
 fun main() {
-    application {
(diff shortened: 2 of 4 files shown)


composeApp/src/commonMain/resources/assets/.gitignore
```diff
@@ -1,2 +0,0 @@
-
-descriptors.json
```
sdsantos, line 2: Why this change?

composeApp/src/desktopMain/kotlin/org/ooni/probe/Main.kt
```diff
@@ -12,20 +13,28 @@ import io.github.vinceglb.autolaunch.AutoLaunch
...
 
 fun main() {
     application {
         val autoLaunch = AutoLaunch(appPackageName = "org.openobservatory.ooniprobe")
 
         var isWindowVisible by remember { mutableStateOf(!autoLaunch.isStartedViaAutostart()) }
+        val deepLinkFlow = MutableSharedFlow<DeepLink?>(extraBufferCapacity = 1)
+        val deepLink by deepLinkFlow.collectAsState(null)
 
+        Desktop.getDesktop().setOpenURIHandler { event ->
```
sdsantos, line 35: `val autoLaunch`, `val deepLinkFlow` and `setOpenURIHandler` should be called outside of `application { }`. Because inside they will run again if there's any re-composition.

sdsantos (approved):

PR #655, opened 2025-04-03
feat: only auto-update descriptors if installed

aanorbel:
Closes https://github.com/ooni/probe-multiplatform/issues/653
The application now checks if test descriptors are installed before attempting to auto-update them.

### DIFF
--- a/composeApp/src/commonMain/kotlin/org/ooni/probe/App.kt
+++ b/composeApp/src/commonMain/kotlin/org/ooni/probe/App.kt
@@ -98,8 +98,10 @@ fun App(
     LaunchedEffect(Unit) {
         dependencies.bootstrapTestDescriptors()
         dependencies.bootstrapPreferences()
-        dependencies.configureDescriptorAutoUpdate()
-        dependencies.startDescriptorsUpdate(null)
+        if (dependencies.hasTestDescriptorInstalled()) {
+            dependencies.configureDescriptorAutoUpdate()
+            dependencies.startDescriptorsUpdate(null)
+        }
         dependencies.startSingleRunInner(RunSpecification.OnlyUploadMissingResults)
     }
     LaunchedEffect(Unit) {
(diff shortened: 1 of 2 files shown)


composeApp/src/commonMain/kotlin/org/ooni/probe/App.kt
```diff
@@ -98,8 +98,10 @@ fun App(
     LaunchedEffect(Unit) {
         dependencies.bootstrapTestDescriptors()
         dependencies.bootstrapPreferences()
-        dependencies.configureDescriptorAutoUpdate()
-        dependencies.startDescriptorsUpdate(null)
+        if (dependencies.hasTestDescriptorInstalled()) {
+            dependencies.configureDescriptorAutoUpdate()
+            dependencies.startDescriptorsUpdate(null)
+        }
```
sdsantos, line 104: This approach works for the app start update. But I feel this isn't the best approach for scheduling. If a user installs a link, but then deletes it at a later stage, it will continue with the daily background updates. I think it's better to match `observeAndConfigureAutoRun()`: continuously observe if there are any links installed, and cancel or schedule the background tasks.

Maybe the `startDescriptorsUpdate` call can also be done inside the observe domain class?

sdsantos (changes requested):

composeApp/src/commonMain/kotlin/org/ooni/probe/App.kt
aanorbel: Ah i see. Let me do that.

composeApp/src/commonMain/kotlin/org/ooni/probe/domain/ObserveAndConfigureAutoUpdate.kt
```diff
@@ -0,0 +1,28 @@
...
+
+class ObserveAndConfigureAutoUpdate(
+    private val backgroundContext: CoroutineContext,
+    private val listAllInstalledTestDescriptors: () -> Flow<List<InstalledTestDescriptorModel>>,
+    private val configureDescriptorAutoUpdate: suspend () -> Boolean,
+    private val cancelDescriptorAutoUpdate: suspend () -> Boolean,
+    private val startDescriptorsUpdate: suspend (List<InstalledTestDescriptorModel>?) -> Unit,
+) {
+    operator fun invoke() =
+        listAllInstalledTestDescriptors()
```
sdsantos, line 18: Just a small thing. This will trigger a new value on every change to descriptors. If you make it like this, it will trigger only when we need to:
```
listAllInstalledTestDescriptors()
   .map { it.isNotEmpty() }
   .distinctUntilChanged()
   .onEach { enabled -> ...
```

sdsantos (approved):

sdsantos (approved):

PR #669, opened 2025-04-10
feat(linux): add deep link handling and new dependency for unique4j

aanorbel:
Closes https://github.com/ooni/probe-multiplatform/issues/640

Reference https://github.com/hydraulic-software/conveyor/issues/5


composeApp/src/desktopMain/kotlin/org/ooni/probe/Main.kt
```diff
@@ -21,16 +21,27 @@ import ooniprobe.composeapp.generated.resources.app_name
...
 
     val deepLinkFlow = MutableSharedFlow<DeepLink?>(extraBufferCapacity = 1)
 
-    Desktop.getDesktop().setOpenURIHandler { event ->
-        deepLinkFlow.tryEmit(DeepLink.AddDescriptor(event.uri.path.split("/").last()))
+    // Initialize the deep link handler
+    val deepLinkHandler = DeepLinkHandler()
+    deepLinkHandler.initialize(args)
+
+    if ((dependencies.platformInfo.platform as? Platform.Desktop)?.os == DesktopOS.Mac) {
```
sdsantos, line 41: Do we also need `registerLinuxProtocolHandler()` for mac? Because right now it looks to be running too no?

composeApp/src/desktopMain/kotlin/org/ooni/probe/Main.kt
```diff
@@ -21,16 +21,27 @@ import ooniprobe.composeapp.generated.resources.app_name
...
+import org.ooni.probe.shared.DesktopOS
+import org.ooni.probe.shared.Platform
 import java.awt.Desktop
 
-fun main() {
-    val autoLaunch = AutoLaunch(appPackageName = "org.openobservatory.ooniprobe")
+val APP_ID = "org.openobservatory.ooniprobe"
+
+fun main(args: Array<String>) {
+    val autoLaunch = AutoLaunch(appPackageName = APP_ID)
```
sdsantos, line 33: Is it ok to run `AutoLaunch` for every instance started? Or should this be after the Unique4j call?

composeApp/src/desktopMain/kotlin/org/ooni/probe/data/DeepLinkHandler.kt
```diff
@@ -0,0 +1,278 @@
...
+                // Register as a protocol handler if not already registered
+                registerProtocolHandler()
+            } else {
+                Logger.d("Could not acquire lock - this should not happen")
+            }
+
+            // Register shutdown hook to free the lock when application exits
+            Runtime.getRuntime().addShutdownHook(
+                Thread {
+                    unique?.freeLock()
```
sdsantos, line 85: Did you check how this handles hard crashes? If we can still open the app? If it uses a lock file, it could still be there.

composeApp/src/desktopMain/kotlin/org/ooni/probe/data/DeepLinkHandler.kt
```diff
@@ -0,0 +1,278 @@
...
+                    val params = parseDeepLink(deepLink)
+                    // messageListeners.forEach { it("Initial deep link: $deepLink, params: $params") }
+                    handleDeepLinkCommand(deepLink, params)
+                }
+
+                // TODO: Implement with conveyor
+                // Register as a protocol handler if not already registered
+                registerProtocolHandler()
+            } else {
+                Logger.d("Could not acquire lock - this should not happen")
```
sdsantos, line 79: Isn't this what happens with the second instance? And should we ignore the args from the second instance? Would a deeplink work for an already existing app?

composeApp/src/desktopMain/kotlin/org/ooni/probe/data/DeepLinkHandler.kt
```diff
@@ -0,0 +1,278 @@
...
+    }
+
+    fun parseDeepLink(deepLink: String): Map<String, String> {
+        val params = mutableMapOf<String, String>()
+
+        if (deepLink.startsWith("ooni://")) {
+            val parts = deepLink.removePrefix("ooni://").split("/")
+            if (parts.size >= 2 && parts[0] == "runv2") {
+                params["command"] = "runv2"
+                params["testId"] = parts[1]
```
sdsantos, line 27: We could already be using the `DeepLink.AddDescriptor` class here.

composeApp/src/desktopMain/kotlin/org/ooni/probe/data/DeepLinkHandler.kt
```diff
@@ -0,0 +1,278 @@
...
+import co.touchlab.kermit.Logger
+import org.ooni.probe.APP_ID
+import tk.pratanumandal.unique4j.Unique4j
+import tk.pratanumandal.unique4j.exception.Unique4jException
+import java.io.File
+import java.nio.file.Files
+import java.nio.file.Paths
+import java.util.Date
+
+class DeepLinkHandler {
```
sdsantos, line 12: I feel we could simplify this class by splitting it into the 3 things it's trying to do:
- Make sure only one app instance is running, and passing the args of every open to the first instance
- Parsing deeplinks
- Registering protocol handlers

composeApp/src/desktopMain/kotlin/org/ooni/probe/data/DeepLinkHandler.kt
aanorbel: The deeplink works if there is an instance already. 

The lock is for unique4j to know the status

composeApp/src/desktopMain/kotlin/org/ooni/probe/data/DeepLinkHandler.kt
aanorbel: I have to check this. I havent had a broken instance if the app crashed at runtime before.

sdsantos:
@aanorbel there are still some unaddressed comments from my last review.

composeApp/src/desktopMain/kotlin/org/ooni/probe/Main.kt
aanorbel: `Unique4j` only applies to `windows` and `linux`  we should allow `AutoLaunch`  as it is for now. No problem encountered the way it is set up.

PR #684, opened 2025-04-15
Cancel run after amount of tests with no internet

### LINKED ISSUE #633: Handle loss of network during a test run
We should stop a test run after a few tests with no valid network.

sdsantos:
Closes #633


aanorbel (approved):
🐳

PR #691, opened 2025-04-22
chore: ensure session close when performing engine operations

aanorbel:
Closes https://github.com/ooni/probe/issues/2853

### DIFF
--- a/composeApp/src/commonMain/kotlin/org/ooni/engine/Engine.kt
+++ b/composeApp/src/commonMain/kotlin/org/ooni/engine/Engine.kt
@@ -7,6 +7,7 @@ import kotlinx.coroutines.flow.channelFlow
 import kotlinx.coroutines.flow.flowOn
 import kotlinx.coroutines.isActive
 import kotlinx.serialization.json.Json
+import okio.use
 import org.ooni.engine.OonimkallBridge.SubmitMeasurementResults
 import org.ooni.engine.models.EnginePreferences
 import org.ooni.engine.models.NetworkType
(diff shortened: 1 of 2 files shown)


composeApp/src/commonMain/kotlin/org/ooni/engine/Engine.kt
```diff
@@ -86,7 +86,16 @@ class Engine(
...
+            val session = session(sessionConfig)
+            try {
+                session.submitMeasurement(measurement)
+            } finally {
+                try {
+                    session.close()
+                } catch (e: Exception) {
+                    Logger.w("Error closing session", e)
+                }
+            }
```
bassosimone, line 98: It feels sad that there's no `defer` statement in Kotlin :disappointed:

Lanius-collaris: Why not use `.use()` method?
https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.io/use.html

bassosimone: Ah, nice, I didn't know about this! :pray:

composeApp/src/commonMain/kotlin/org/ooni/engine/OonimkallBridge.kt
```diff
@@ -47,8 +49,7 @@ interface OonimkallBridge {
         fun httpDo(request: HTTPRequest): HTTPResponse
 
         // Needs to be called after `checkIn`
```
sdsantos, line 51: No need for this comment now, we need it for every use of session it seems.

composeApp/src/commonMain/kotlin/org/ooni/engine/Engine.kt
```diff
@@ -86,17 +87,18 @@ class Engine(
...
         }.mapError { MkException(it) }
 
     suspend fun checkIn(taskOrigin: TaskOrigin): Result<OonimkallBridge.CheckInResults, MkException> {
         return resultOf(backgroundContext) {
             val preferences = getEnginePreferences()
             val sessionConfig = buildSessionConfig(taskOrigin, preferences)
-            val session = session(sessionConfig)
-            try {
+            session(sessionConfig).run {
```
sdsantos, line 99: It should be `use` and not `run`.

sdsantos (changes requested):

PR #704, opened 2025-05-05
Desktop app icon

### LINKED ISSUE #698: desktop: App icons

sdsantos:
Closes #698 

Conveyor automatically generates the appropriate icon format for each platform, based on the SVG: https://conveyor.hydraulic.dev/15.1/configs/icons/


icons/app.svg
aanorbel, line 1: i think we should go with a *Squircle*

aanorbel (approved):
:whale:

icons/app.svg
sdsantos: Is that more common in Windows as well? Or just mac? We can have different ones for each platform too.

aanorbel: Squircle is almost a must for Mac. For Windows and Linux, we can get away with the circular icon.

sdsantos: But then we could use Squircle for all platforms, what do you think?

sdsantos:
New version

Screenshot 2025-05-06 at 12 27 35

PR #707, opened 2025-05-05
fix: onboarding in tablet landscape

aanorbel:


sdsantos:
I don't think this is ideal still. I can try some middle-ground solution tomorrow, to see how it looks.

Screenshot 2025-05-05 at 18 49 35
Screenshot 2025-05-05 at 18 50 20

PR #717, opened 2025-05-09
feat: `add-descriptor` webconnectivity urls

aanorbel:
Closes https://github.com/ooni/probe-multiplatform/issues/710

### DIFF
--- a/composeApp/src/commonMain/kotlin/org/ooni/probe/ui/descriptor/add/AddDescriptorScreen.kt
+++ b/composeApp/src/commonMain/kotlin/org/ooni/probe/ui/descriptor/add/AddDescriptorScreen.kt
@@ -5,7 +5,11 @@ import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.WindowInsets
+import androidx.compose.foundation.layout.asPaddingValues
+import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.navigationBars
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.lazy.LazyColumn
 import androidx.compose.foundation.lazy.items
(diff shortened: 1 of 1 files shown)


composeApp/src/commonMain/kotlin/org/ooni/probe/ui/descriptor/add/AddDescriptorScreen.kt
```diff
@@ -131,18 +136,37 @@ fun AddDescriptorScreen(
                         modifier = Modifier.weight(1f),
                     )
                 }
-                LazyColumn {
+            }
+
+            Box(
```
sdsantos, line 141: Why is this Box needed?

composeApp/src/commonMain/kotlin/org/ooni/probe/ui/descriptor/add/AddDescriptorScreen.kt
```diff
@@ -131,18 +136,37 @@ fun AddDescriptorScreen(
...
+                LazyColumn(modifier = Modifier.padding(horizontal = 16.dp).padding(bottom = 48.dp)) {
                     items(state.selectableItems) { selectableItem ->
                         TestItem(selectableItem, onChecked = { _ ->
                             onEvent(
                                 AddDescriptorViewModel.Event.SelectableItemClicked(selectableItem),
                             )
                         })
+
+                        if (state.selectableItems.size == 1 && selectableItem.item.test == TestType.WebConnectivity) {
+                            selectableItem.item.inputs.orEmpty().forEach { website ->
```
sdsantos, line 155: This needs to be a LazyColumn, since lists can have thousands of links. So ideally TestItems and URLs should be sharing the LazyColumn above-

sdsantos (changes requested):
I feel the URLs should either align with the checkbox or the "Web Connectivity Test". Like this it feels a bit off:

Screenshot 2025-05-12 at 11 52 44

composeApp/src/commonMain/kotlin/org/ooni/probe/ui/descriptor/add/AddDescriptorScreen.kt
aanorbel: Adding the box to ensure the actions are displayed at the bottom of the screen and not at the end of the URL list.
This ensures we can install the descriptor without having to scroll to the end of the list when its a long one.

sdsantos (approved):

Issue #660, opened 2025-04-07
desktop: implement NetworkTypeFinder using native APIs


sdsantos:
From @hellais 

I wasn't able to readily find some library that supports this in a multiplatform way, but maybe doing some searches with those OS libraries as primitives could yield some results.

For macOS it's worth looking into what this library is doing: https://github.com/ashleymills/Reachability.swift/blob/master/Sources/Reachability.swift
Though it's using an API that apparently has been deprecated: https://developer.apple.com/documentation/systemconfiguration/scnetworkreachability-g7d
If we are targeting macOS 10.14+, it seems like the new set of libraries to use is this: https://developer.apple.com/documentation/Network/NWPathMonitor
Which has this type definition: https://developer.apple.com/documentation/network/nwinterface/interfacetype
which can be accessed from this struct: https://developer.apple.com/documentation/network/nwinterface

On windows, if we only care about Window 10+, then this looks like the API we need: https://learn.microsoft.com/en-us/uwp/api/windows.networking.connectivity.connectionprofile?view=winrt-26100&redirectedfrom=MSDN

PR #729, opened 2025-05-24
feat: implement network type detection using JNI

aanorbel:
Closes https://github.com/ooni/probe-multiplatform/issues/660

Check proper embedding of lib with conveyor


composeApp/src/desktopMain/kotlin/org/ooni/probe/BuildDependencies.kt
```diff
@@ -37,7 +37,7 @@ val dependencies = Dependencies(
     cacheDir = projectDirectories.cacheDir.also { File(it).mkdirs() },
     readAssetFile = ::readAssetFile,
     databaseDriverFactory = { buildDatabaseDriver(projectDirectories.dataDir) },
-    networkTypeFinder = ::networkTypeFinder,
+    networkTypeFinder = MacOsNetworkTypeFinder(),
```
sdsantos, line 40: We probably should only use `MacOsNetworkTypeFinder` for macos.

sdsantos (changes requested):
Just missing making the library during the build step.

PR #736, opened 2025-05-27
Mark all results as viewed

sdsantos:
Toolbar button to mark all results as viewed.

@hellais and @agrabeli there's more copy to review:

`Do you want to mark all results as viewed?`


agrabeli:
Thanks @sdsantos LGTM!

PR #743, opened 2025-06-02
Descriptor run button

### LINKED ISSUE #741: Run button on descriptor screen
Descriptors are listed prominently in our home screen. And although we have a big Run button at the top, users might still expect to be able to Run a specific test when opening a descriptor of that test. So I suggest we place a Run button there too. It would open the same Run screen, with just the tests of that descriptor selected.

Together with this we should make another change. Right now, inside the Run screen, when the user selects/deselects tests to run, we automatically save that configuration in the preferences for next time. We can instead only do it if the user actually presses “Run” for that configuration. That would avoid the user losing its default confirmation just by tapping the Run button on the Descriptor screen.

Image

sdsantos:
Closes #741

---

https://github.com/user-attachments/assets/247f8999-f397-4891-9338-65ec9de20119


aanorbel (approved):
🐳

PR #754, opened 2025-06-09
Results Date filter

### LINKED ISSUE #746: Filter results by date
You can filter results by manual/auto run, and by test descriptor. But you can’t filter by a date interval, for example to know if anything was blocked on a specific day/week, a couple of months ago. You need to scroll a lot for that, and it’s only visible if it’s on the more recent 100 results.

Implementation suggestion:
- A small icon button with a calendar after “All Sources”
- Tapping the button would open a dialog to pick a date interval
- If an interval is selected, a new line is added below “Filter Tests” with the current date intervale
- Tapping the data interval would reopen the dialog to change the interval
- There would also be a clear button next to the date interval, to clear the current date filter.

Image

sdsantos:
Closes #746 

@hellais @agrabeli there are new strings to validate in these screens:

### DIFF
--- a/composeApp/src/commonMain/kotlin/org/ooni/probe/data/models/ResultFilter.kt
+++ b/composeApp/src/commonMain/kotlin/org/ooni/probe/data/models/ResultFilter.kt
@@ -9,7 +9,7 @@ import org.ooni.probe.shared.today
 data class ResultFilter(
     val descriptors: List<Descriptor> = emptyList(),
     val taskOrigin: TaskOrigin? = null,
-    val dates: Date = Date.Any,
+    val dates: Date = Date.AnyDate,
     val limit: Long = LIMIT,
 ) {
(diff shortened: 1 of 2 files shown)


composeApp/src/commonMain/kotlin/org/ooni/probe/data/models/ResultFilter.kt
```diff
@@ -1,14 +1,37 @@
...
     val taskOrigin: TaskOrigin? = null,
+    val dates: Date = Date.Any,
     val limit: Long = LIMIT,
 ) {
     val isAll get() = this == ResultFilter()
 
+    sealed class Date(
+        val range: ClosedRange<LocalDate>,
+    ) {
+        data object Any :
```
aanorbel, line 20: We should probably rename this to revent confusion in the future

aanorbel (approved):
:whale:

PR #755, opened 2025-06-09
Delete results and mark as viewed by filter

### LINKED ISSUE #747: Delete results based on current filter
Currently, you can only delete all results in the results screen if no filter is selected. You can’t delete all results of a specific descriptor, or all manual or auto runs. We could implement that functionality.

sdsantos:
Closes #747


aanorbel (approved):
🐳  LGTM

Issue #748, opened 2025-06-03
Selective results deletion


sdsantos:
Right now you can only delete all the results you have stored on your device. You can’t just delete one or more specific results. Imagine some results failed because of a lack of Internet connection. Or you're worried about someone seeing a specific result on your phone.

You could long press to select one or multiple results, and then have the ability to delete the selected results.

PR #756, opened 2025-06-09
feat: select results ui

aanorbel:
Closes https://github.com/ooni/probe-multiplatform/issues/748

### DIFF
--- a/composeApp/src/commonMain/kotlin/org/ooni/probe/ui/results/ResultsScreen.kt
+++ b/composeApp/src/commonMain/kotlin/org/ooni/probe/ui/results/ResultsScreen.kt
@@ -13,6 +13,8 @@ import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.lazy.LazyColumn
 import androidx.compose.foundation.lazy.items
+import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.automirrored.filled.ArrowBack
 import androidx.compose.material3.AlertDialog
 import androidx.compose.material3.CircularProgressIndicator
 import androidx.compose.material3.HorizontalDivider
@@ -23,6 +25,9 @@ import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Surface
 import androidx.compose.material3.Text
 import androidx.compose.material3.TextButton
+import androidx.compose.material3.TopAppBar
+import androidx.compose.material3.TopAppBarDefaults
+import androidx.compose.material3.TriStateCheckbox
 import androidx.compose.material3.VerticalDivider
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
@@ -35,6 +40,7 @@ import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.alpha
 import androidx.compose.ui.draw.shadow
 import androidx.compose.ui.platform.LocalDensity
+import androidx.compose.ui.state.ToggleableState
 import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.unit.dp
 import kotlinx.datetime.LocalDate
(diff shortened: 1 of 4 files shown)


composeApp/src/commonMain/kotlin/org/ooni/probe/ui/results/ResultsViewModel.kt
```diff
@@ -163,5 +213,15 @@ class ResultsViewModel(
...
+
+        data class SelectResult(val id: ResultModel.Id) : Event
+
+        data class DeselectResult(val id: ResultModel.Id) : Event
+
+        data object CancelSelection : Event
+
+        data object ToggleSelection : Event
+
+        data object DeleteSelectedResults : Event
```
sdsantos, line 225: We could reuse the DeleteAllClick, rename it into a DeleteClick, and use according to the selection state.

composeApp/src/commonMain/kotlin/org/ooni/probe/ui/results/ResultsViewModel.kt
```diff
@@ -127,11 +173,15 @@ class ResultsViewModel(
         val summary: Summary? = null,
         val isLoading: Boolean = true,
         val markAllAsViewedEnabled: Boolean = false,
+        val selectedResults: Set<ResultModel.Id> = emptySet(),
```
sdsantos, line 176: You could also have wrapped the ResultListItem with `SelectableItem` like we've done in other screens. But you would still need to keep the state inside the ViewModel

composeApp/src/commonMain/kotlin/org/ooni/probe/ui/results/ResultsViewModel.kt
```diff
@@ -127,11 +173,15 @@ class ResultsViewModel(
...
+        val selectionEnabled: Boolean = false,
     ) {
         val anyMissingUpload
             get() = results.any { it.value.any { item -> !item.allMeasurementsUploaded } }
 
         val areResultsLimited get() = results.values.sumOf { it.size } >= ResultFilter.LIMIT
+        val isAllSelected get() = selectedResults.size == results.values.sumOf { it.size }
```
sdsantos, line 183: Small english thing: `isAllSelected` -> `areAllSelected`

sdsantos (changes requested):
UI:

a) Status bar color needs to match the background.
b) Don't show the Needs Upload bar when selecting. It may seem that you are just uploading the selected results.
c) Select all tests needs to be clickable, not just the checkbox. Use the `triStateToggleable` modifier.
d) "Select all" should become "Deselect all" if all are selected
e) "Select all tests" should be "Select all test results" or "Select all results"
f) If I press back, the results say selected in the list
g) If no results are selected, the delete button could be disabled instead of disappearing, so users always know what actions they can do

composeApp/src/commonMain/kotlin/org/ooni/probe/ui/results/ResultsViewModel.kt
```diff
@@ -85,8 +93,30 @@ class ResultsViewModel(
...
+                    }
+                    if (selectedIds.isNotEmpty()) {
+                        deleteResults(selectedIds)
+                        // Deselect all after deletion
+                        _state.update {
+                            it.copy(
+                                results = it.results.mapValues {
+                                        (_, list) ->
+                                    list.map { item -> item.copy(isSelected = false) }
+                                },
```
sdsantos, line 113: You shouldn't have to do this, because those items will be deleted and removed from the list. You just need to turn off selection.

composeApp/src/commonMain/kotlin/org/ooni/probe/ui/results/ResultsScreen.kt
```diff
@@ -95,6 +97,8 @@ fun ResultsScreen(
     state: ResultsViewModel.State,
     onEvent: (ResultsViewModel.Event) -> Unit,
 ) {
+    LightStatusBars(state.selectionEnabled && MaterialTheme.colorScheme.background.luminance() > 0.5f)
```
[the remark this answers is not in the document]
aanorbel: if you check the summary, all the app uses light status bar except this particular case.

PR #757, opened 2025-06-10
Fix dashboard logo

sdsantos:
The OONI Probe logo wasn't appearing in the Dashboard screen, in portrait. We had an extra statusBars inset padding. I also reduced a bit the height of the blue section.


aanorbel (approved):
🐳

PR #759, opened 2025-06-10
Persist chosen websites

sdsantos:


aanorbel (approved):
I think the user should be prompted if they want to persist the urls.

PR #765, opened 2025-06-13
chore: update oonikmall to `v3.26.0`

aanorbel:


composeApp/src/commonMain/kotlin/org/ooni/probe/ui/navigation/BottomNavigationBar.kt
```diff
@@ -48,6 +48,7 @@ fun BottomNavigationBar(
         modifier = customMinHeightModifier,
     ) {
         MAIN_NAVIGATION_SCREENS.forEach { screen ->
+            val screen = screen as Screen
```
sdsantos, line 51: Was lint complaining or something?

aanorbel: iOS was complaining that screen is of type `Any`

PR #773, opened 2025-06-17
fix: properly handle `AS0`

aanorbel:


composeApp/src/commonMain/kotlin/org/ooni/probe/domain/RunNetTest.kt
```diff
@@ -217,8 +218,15 @@ class RunNetTest(
...
-                    measurement.reportFilePath?.let { deleteFiles(it) }
-                    measurement.copy(isUploaded = true)
+                    return@updateMeasurement if (lastNetwork?.isValid() == true) {
+                        measurement.reportFilePath?.let { deleteFiles(it) }
+                        measurement.copy(isUploaded = true)
+                    } else {
+                        measurement.copy(
+                            isUploadFailed = true,
+                            uploadFailureMessage = "Network is not valid",
+                        )
```
sdsantos, line 228: Should we mark the upload as failed, or the whole measurement as failed? Because upload will never work, even if we update the Geo IP tables in the app. And like this it will retry every time.

sdsantos (changes requested):

PR #776, opened 2025-06-18
Use plurals resources directly

### LINKED ISSUE #772: Enhance pluralization support
One of our translations is raising the issue that our current pluralizations only support `one` or `other` and we could be more thorough. The full range is: zero, one, two, few, many, other. (https://www.unicode.org/cldr/charts/47/supplemental/language_plural_rules.html)

We should add them all, and use the Transifex's support for plural resources directly, to simplify things.

sdsantos:
Closes #772 

Transifex should support Android plurals directly, expanding the plurals with other items for languages that need them. But we'll only be able to confirm once we merge it.


aanorbel (approved):
🐳

PR #815, opened 2025-07-07
Add scrollbars on desktop for main screens

### LINKED ISSUE #812: desktop: Show scrollbars on scrollable content
Sometimes it's not clear that there is additional content to scroll on desktop. And not all users have access to a mouse with scroll wheel. So it's better to always show the scrollbars on desktop.

sdsantos:
Closes #812 

There's no general solution. The `VerticalScrollbar` component only exists on the Desktop platform.


aanorbel (approved):
🐳

PR #819, opened 2025-07-08
Update AGP and wrapper

sdsantos:


sdsantos:
@aanorbel looks like the new ktlint introduce a lot of new rules, changing tons of files. Let me know if you agree with the changes, or if we need to disable/change some new rule.

aanorbel (approved):
🐳

Issue #817, opened 2025-07-08
desktop: Detect and suggest deletion of previous version data


sdsantos:

aanorbel:
**macOS**
- **Application Location**: `/Applications/OONI Probe.app`
- **User Data and Configuration**: 
  - Path: `~/Library/Application Support/OONI Probe/`
  - Autorun: `~/Library/LaunchAgents/org.ooni.probe-desktop.plist` (manages background tasks)
- **Logs**:
  - Test Logs: `~/Library/Application Support/OONI Probe/`
  - Additional Logs: `~/Library/Logs/OONI Probe/`
- **Configuration**:
  - OONI Home Config: `~/Library/Application Support/OONI Probe/ooni_home/config.json`

**Windows**
- **Application Location**: `C:\Program Files\OONI Probe\`
- **User Data and Configuration**: 
  - Path: `%APPDATA%\OONI Probe\` (e.g., `C:\Users\<Username>\AppData\Roaming\OONI Probe\`)
- **Logs**: `%APPDATA%\OONI Probe\`
- **Configuration**: `%USERPROFILE%\AppData\Roaming\OONI Probe\`

**Note**: Windows uninstaller reference: https://github.com/ooni/probe-desktop/blob/master/build/windows/uninstaller.nsh

PR #823, opened 2025-07-14
feat(macos): add cleanup step for legacy directories

aanorbel:
Closes https://github.com/ooni/probe-multiplatform/issues/817

Screenshot 2025-07-14 at 16 56 04


sdsantos (changes requested):
Tested it and it didn't delete my Application Support folder `OONI Probe`.

Also, don't forget to ask for a copy review.

composeApp/src/desktopMain/kotlin/org/ooni/probe/BuildDependencies.kt
```diff
@@ -24,6 +24,7 @@ import java.net.URLEncoder
 import java.nio.charset.StandardCharsets
 
 private val projectDirectories = ProjectDirectories.from("org", "OONI", "Probe")
+private val oldProjectDirectories = ProjectDirectories.fromPath("OONI Probe")
```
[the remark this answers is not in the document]
aanorbel: I checked on macos. Checking on windows now.

composeApp/src/ooniMain/kotlin/org/ooni/probe/ui/onboarding/OnboardingUtils.kt
```diff
@@ -0,0 +1,54 @@
...
+import ooniprobe.composeapp.generated.resources.Common_Clear
+import ooniprobe.composeapp.generated.resources.Res
+import org.ooni.probe.config.OrganizationConfig
+
+import ooniprobe.composeapp.generated.resources.Onboarding_CleanUp_Paragraph
+import ooniprobe.composeapp.generated.resources.Onboarding_CleanUp_Title
+import ooniprobe.composeapp.generated.resources.Onboarding_Crash_Button_No
+
+@Composable
+fun ColumnScope.CleanUpStep(onEvent: (OnboardingViewModel.Event) -> Unit) {
```
sdsantos, line 24: This could have been implemented for both platforms, I don't see harm in it. That would lower the amount of code for this feature, and keep more functions private.

composeApp/src/desktopMain/kotlin/org/ooni/probe/BuildDependencies.kt
```diff
@@ -48,11 +74,12 @@ val dependencies = Dependencies(
     launchAction = ::launchAction,
     batteryOptimization = object : BatteryOptimization {},
     isWebViewAvailable = { true },
+    isCleanUpRequired = ::hasLegacyDirectories,
+    cleanupLegacyDirectories = ::cleanupLegacyDirectories,
```
sdsantos, line 78: This methods could be on a separate class/file, to continuously growing the BuildDependencies file. Also, they could be bundled together like on Battery Optimization.

composeApp/src/desktopMain/kotlin/org/ooni/probe/BuildDependencies.kt
```diff
@@ -173,6 +200,51 @@ private fun buildMailUri(action: PlatformAction.Mail): URI {
...
+        deletePath(dirPath).also {
+            if (it) {
+                Logger.i { "Successfully cleaned up legacy path: $dirPath" }
+            } else {
+                Logger.w { "Failed to clean up legacy path: $dirPath" }
+            }
+        }
+    }
+    Logger.i { "Legacy directory cleanup process finished." }
+    return results.all { it }
```
sdsantos, line 223: We're not using this value, at least for now.

composeApp/src/ooniMain/kotlin/org/ooni/probe/ui/onboarding/OnboardingUtils.kt
aanorbel: DW has no previous versions per say

composeApp/src/desktopMain/kotlin/org/ooni/probe/shared/LegacyDirectoryManager.kt
```diff
@@ -0,0 +1,81 @@
...
+        else -> emptyList()
+    }
+
+    fun hasLegacyDirectories(): Boolean =
+        legacyPaths.any { path ->
+            val file = File(path)
+            file.exists() && file.isDirectory && file.listFiles().any()
+        }
+
+    suspend fun cleanupLegacyDirectories(): Boolean {
```
sdsantos, line 43: I remembered that this should be running on a background CoroutineContext. You should pass the backgroundContext as an argument and wrap the body of this method with `withContext(...) { ... }`.

composeApp/src/ooniMain/kotlin/org/ooni/probe/ui/onboarding/OnboardingUtils.kt
sdsantos: I know, that's why I say it doesn't matter much. But I guess the onboarding image is OONI specific, so let's leave it.

sdsantos (changes requested):
Just to final comments, but they are important things to change.

sdsantos (approved):
Not a big deal, but after I clear the leftover files in Settings, the option should disappear, if it was successful. Maybe create a Flow for checking if there are leftover files, add it to the `combine` on top of the `GetSettings`, and trigger it after `cleanupLegacyDirectories?.invoke()`.

PR #825, opened 2025-07-15
Fix navigation crash

sdsantos:
Fixes this crash: https://ooni.sentry.io/issues/6716393476/
And also a bug where the Running Screen, at the end of a test run, wasn't returning successfully to the results screen.


aanorbel (approved):
🐳

PR #838, opened 2025-07-22
Remove some build warnings and prints

sdsantos:
I think we can remove the most frequent prints from the build, and just leave the ones that may point to issues.


aanorbel (approved):
🐳

PR #878, opened 2025-09-04
fear(windows): add deeplink support

aanorbel:


composeApp/src/desktopMain/kotlin/org/ooni/probe/Main.kt
```diff
@@ -142,6 +144,37 @@ fun main(args: Array<String>) {
                 )
             },
         )
+
+        LaunchedEffect(Unit) {
```
sdsantos, line 148: Some questions:
- Is it ok to this run this on every app start?
- This will run on the main thread, could it run on the background?
- And could this be inside a function, to avoid having a longer `main`?

PR #879, opened 2025-09-04
chore: remove Conveyor plugin and related configurations

aanorbel:


.github/workflows/desktop_make.yml
```diff
@@ -15,20 +15,8 @@ jobs:
       - name: Package app
         run: ./gradlew copyBrandingToCommonResources package -Porganization=ooni
```
sdsantos, line 16: Now this will only build for the current platform right? We need to run this on 2 or 3 different machines to get our binaries.

sdsantos (approved):

PR #905, opened 2025-09-16
Fix wrong translatable=false

sdsantos:
Found a couple of plurals that should have been translated, but were wrongfully marked as not for translation.


aanorbel (approved):
🐳  LGTM

PR #911, opened 2025-09-17
feat(windows): add support for `winsparkle` updates

aanorbel:


composeApp/src/desktopMain/kotlin/org/ooni/probe/update/DesktopUpdateController.kt
```diff
@@ -0,0 +1,128 @@
...
+    fun getMenuText(): String =
+        when {
+            _error.value != null -> "Retry Update Check (Error: ${_error.value?.code})"
+            _state.value == UpdateState.CHECKING_FOR_UPDATES -> "Checking for Updates..."
+            _state.value == UpdateState.UPDATE_AVAILABLE -> "Update Available!"
+            _state.value == UpdateState.NO_UPDATE_AVAILABLE -> "Check for Updates (Up to date)"
+            _state.value == UpdateState.INITIALIZING -> "Initializing Updates..."
+            _state.value == UpdateState.ERROR -> "Update System Error - Retry"
+            else -> "Check for Updates..."
+        }
```
sdsantos, line 117: We need to localize all these messages, either here, or in the UI code. Could be in a separate ticket.

composeApp/src/desktopMain/kotlin/org/ooni/probe/update/DesktopUpdateController.kt
```diff
@@ -0,0 +1,128 @@
...
+
+    suspend fun cleanup() {
+        try {
+            updateManager.cleanup()
+            Logger.i("Update manager cleaned up successfully")
+        } catch (e: Exception) {
+            Logger.e("Error during update manager cleanup: $e")
+            throw e
```
sdsantos, line 125: Why are we handling it here, if we're re-throwing the exception to be caught the same way by the caller? Either not handle it, and add a `@Throws` annotation, or handle it here and not re-throw it.

sdsantos (approved):

PR #914, opened 2025-09-17
Descriptor screen run button starts tests immediately

### LINKED ISSUE #909: Pressing Run on the Descriptor screen skips the Run screen
When pressing the Run button inside the Descriptor screen, it skips the Run screen, and starts a manual test immediately with just that descriptor (including all its tests).

It should check if the VPN warning is required, just like the Run screen.

sdsantos:
Closes #909


aanorbel (approved):
🐳  LGTM

PR #919, opened 2025-09-24
New proxy screen

### LINKED ISSUE #416: desktop, cli, mobile: tell users about bad proxies
This issue is about telling users that they have configured a bad proxy. For example, suppose `tor` is listening at `socks5://127.0.0.1:9050/` and I mistype it as `socks5://127.0.0.2:9050/`, it would be nice for me, as a user, to know that the OONI Probe engine cannot boostrap because of such a typo.

There are two obvious approaches here. The first approach is wrapping proxy errors such that they are recognizable by the specific OONI Probe App. The second approach is testing the proxy right after it is configured.

The problem with the first approach is that "all available probe services failed", the error that is returned when the proxy is not working, is a composed error containing an error for each attempt. Therefore, to conclude that we have a proxy error, we would need somehow to aggregate to conclude something like "proxy error: all available probe services failed".

The problem with testing a proxy right after it has been configured is that a proxy may work at time T1 but may stop working at time T2. For example, I may be using Orbot and Orbot may be stopped some time in the future. So, when I configure the proxy at time T1 all looks good, but we subsequently see a failure at time T2 when I use it.

An additional approach could be for the OONI Engine to test the proxy before using it, when it is configured to use a proxy. If we factor in the case of tor, torsf, and psiphon, this would mean testing the proxy after the tor, torsf, or psiphon bootstrap.

If we go down with this approach, we would have a preliminary step, which consists of checking whether we can use the proxy and which fails with an informative error in case the proxy does not work. Obviously, this means that, if the proxy fails at a later time, we will have several failures, but at least we can say it was working when we started measuring.

sdsantos:
Closes #416 

https://github.com/user-attachments/assets/41a9cec3-afb9-4adb-adad-4e9ae90fe6ba


aanorbel (changes requested):
Everything else LGTM

PR #933, opened 2025-09-30
Improve run message for single web connectivity tests

### LINKED ISSUE #932: Display the descriptor title when running a single Web Connectivity descriptor
- In the running message on the home screen, if we're running a single Web Connectivity test descriptor, the “Running:” message should display the descriptor title instead of "Web Connectivity Test".
- The Running screen also should show the title of descriptor, which doesn't at the moment.

sdsantos:
Closes #932


aanorbel (approved):
🐳

PR #938, opened 2025-10-01
Make descriptor sections collapsable

### LINKED ISSUE #931: Make the descriptor sections collapsable
We should store in the preferences weather a section is collapsed or expanded, so the state is kept across sessions

sdsantos:
Closes #931 

Also found a bug where the descriptors didn't update on desktop, when we wanted to check for all descriptors, and not one in particular. So now we handle the empty descriptor list inside `FetchDescriptorsUpdates`.

|.|.|
|-|-|
|Screenshot 2025-10-01 at 17 29 41|Screenshot 2025-10-01 at 17 29 51|


aanorbel (approved):
🐳

PR #945, opened 2025-10-07
Fix lottie animation during onboarding

sdsantos:
It never showed the animation, because `onFinish` was being called straight away.


aanorbel (approved):
🐳

PR #976, opened 2025-10-28
Run tests in new thread

sdsantos:
Run tests explicitly in a new thread, to see if it helps with handling native crashes. I also added error handling per individual test, so one test throwing an exception does not stop the entire run.


aanorbel (approved):
🐳

PR #989, opened 2025-11-04
Show list of countries

### LINKED ISSUE #979: Show list of countries already tested
This can be shown by tapping on the countries number on the new dashboard stats

sdsantos:
Closes #979 

It's a small Easter egg for users.

I used each platform method for getting the country names, instead of using our own list of countries, that would have to be translated. The downside is that the country names don't respond to composable changes (change of device language), unless the stats values change too.

https://github.com/user-attachments/assets/9e7edb29-88c2-456d-9fde-517c1f138487

### DIFF
--- a/composeApp/src/desktopMain/kotlin/org/ooni/probe/BuildDependencies.kt
+++ b/composeApp/src/desktopMain/kotlin/org/ooni/probe/BuildDependencies.kt
@@ -171,7 +171,13 @@ fun sendMail(action: PlatformAction.Mail): Boolean =
         false
     }
 
-private fun getCountryNameByCode(countryCode: String) = Locale("", countryCode).displayCountry.ifEmpty { countryCode }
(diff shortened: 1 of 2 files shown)


composeApp/src/desktopMain/kotlin/org/ooni/probe/BuildDependencies.kt
```diff
@@ -169,6 +171,8 @@ fun sendMail(action: PlatformAction.Mail): Boolean =
         false
     }
 
+private fun getCountryNameByCode(countryCode: String) = Locale("", countryCode).displayCountry.ifEmpty { countryCode }
```
aanorbel:
```suggestion
private fun getCountryNameByCode(countryCode: String) = Locale.Builder().setRegion(countryCode).build().displayCountry.ifEmpty { countryCode }
```

PR #1036, opened 2025-12-15
Smaller APK resources

sdsantos:
The OTF fonts are smaller that the TTFs. And we had the DW onboarding image in common, bundled on every APK.

Saved around 300KB in final APK size.


aanorbel (approved):
🐳

PR #1050, opened 2026-01-12
refactor(models): rename `InstalledTestDescriptorModel` to `Descriptor` and introduce `DescriptorItem` class for UI logic

aanorbel:


composeApp/src/commonMain/kotlin/org/ooni/probe/domain/GetAutoRunSpecification.kt
```diff
@@ -20,30 +20,30 @@ class GetAutoRunSpecification(
             .notExpired()
 
         return RunSpecification.Full(
-            tests = descriptors.map { descriptor ->
-                RunSpecification.Test(
-                    source = descriptor.source.id,
-                    netTests = descriptor.netTests,
-                )
+            tests = descriptors.mapNotNull { descriptor ->
```
sdsantos: Not sure why this code moved from `filterForAutoRun()` to here. Not sure what's different now.

composeApp/src/commonMain/kotlin/org/ooni/probe/data/models/DescriptorItem.kt
```diff
@@ -0,0 +1,158 @@
...
+import org.ooni.engine.models.SummaryType
+import org.ooni.engine.models.TestType
+import org.ooni.probe.shared.InstalledDescriptorIcons
+import org.ooni.probe.shared.hexToColor
+import org.ooni.probe.shared.now
+import org.ooni.probe.shared.stringMonthArrayResource
+import kotlin.time.Duration.Companion.seconds
+
+data class DescriptorItem(
+    val source: Descriptor,
```
sdsantos: Not sure if `source` is more understandable here than just `descriptor`. I guess it can avoid some `descriptor.descriptor`, but ideally we should always have `descriptorItem.descriptor` or `item.descriptor`.

PR #1063, opened 2026-01-16
feat: add listener for app reopening on desktop

### LINKED ISSUE #1049: desktop: Show/hide app from dock
When the app window is visible, the app should be visible on the mac dock and the task switcher. And it should hide from the dock when the window is not visible

### LINKED ISSUE #1058: desktop: Move app to foreground when re-opening app

aanorbel:
Closes #1049 
Closes #1058

### DIFF
--- a/composeApp/src/desktopMain/kotlin/org/ooni/probe/Main.kt
+++ b/composeApp/src/desktopMain/kotlin/org/ooni/probe/Main.kt
@@ -183,15 +183,13 @@ fun main(args: Array<String>) {
         }
 
         LaunchedEffect(Unit) {
-            if (Desktop.isDesktopSupported()) {
-                runCatching {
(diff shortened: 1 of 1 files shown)


composeApp/src/desktopMain/kotlin/org/ooni/probe/Main.kt
```diff
@@ -179,6 +181,19 @@ fun main(args: Array<String>) {
         CoroutineScope(Dispatchers.Default).launch {
             registerWindowsUrlScheme()
         }
+
+        LaunchedEffect(Unit) {
+            if (Desktop.isDesktopSupported()) {
```
sdsantos: Do you know if we need all these checks for desktop is supported? In theory Main.kt is already only for desktop...

PR #1178, opened 2026-03-03
feat: user registration using manifest credentials

aanorbel:
Closes https://github.com/ooni/probe-multiplatform/issues/1159


composeApp/src/commonMain/kotlin/org/ooni/probe/domain/credentials/RegisterUserWithManifest.kt
```diff
@@ -0,0 +1,60 @@
...
+
+                if (publicParameters.isBlank()) {
+                    Logger.w("Manifest has empty publicParameters")
+                    return@withContext null
+                }
+
+                if (manifestVersion.isBlank()) {
+                    Logger.w("Manifest has empty version")
+                    return@withContext null
+                }
```
sdsantos: We can also validate this before storing the Manifest, to avoid having to check after.

sdsantos (approved):

PR #1286, opened 2026-04-10
Optimize Articles and Urls bulk insertOrReplace

sdsantos:
Sentry was complaining about the performance of some of our bulk queries:
- https://ooni.sentry.io/issues/7367486872/
- https://ooni.sentry.io/issues/7348615212/

So I optimized a bit the related operations:
- Minimized the amount of SQL queries on Articles.refresh
- Fixed a bug on the UrlRepository where models were always updates, even when they matched the ones on the DB.
- Simplified a bit the code inside UrlRepository


aanorbel (approved):
🐳

PR #1289, opened 2026-04-10
desktop: Debug build support

### LINKED ISSUE #1281: desktop: Support for debug builds
Currently, all desktop builds run as `release` builds, using the production API. It would be nice to be able to test with `debug` builds like on Android.

sdsantos:
Closes #1281

@aanorbel what do you think of this approach? It works for iOS as well. And we can expand it to change the desktop app name, the folder, etc..

### DIFF
--- a/buildSrc/src/main/kotlin/BuildUtils.kt
+++ b/buildSrc/src/main/kotlin/BuildUtils.kt
@@ -17,13 +17,15 @@ fun Project.isFdroidTaskRequested(): Boolean =
 fun Project.isDebugTaskRequested(): Boolean {
     val isTaskDebug = gradle.startParameter.taskRequests
         .flatMap { it.args }
-        .any { it.contains("Debug") }
+        .any { it.contains("Debug") || it.contains("Run") }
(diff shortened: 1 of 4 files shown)


buildSrc/src/main/kotlin/BuildUtils.kt
```diff
@@ -13,10 +14,22 @@ fun Project.isFdroidTaskRequested(): Boolean =
 /**
  * Check if Debug build task is requested.
  */
-fun Project.isDebugTaskRequested(): Boolean =
-    gradle.startParameter.taskRequests
+fun Project.isDebugTaskRequested(): Boolean {
+    val isTaskDebug = gradle.startParameter.taskRequests
         .flatMap { it.args }
         .any { it.contains("Debug") }
```
aanorbel:
```suggestion
        .any { it.contains("Debug") || it.contains("Run") }
```
Expanded to contain more tasks patterns.

aanorbel (changes requested):
I think this can work.

aanorbel (approved):
🐳

PR #1292, opened 2026-04-13
Handle empty measurementUid responses

sdsantos:
This was happening on the `dev` environment but it's better to protect against this happening. It should fallback to using the reportId for showing the measurement page.


aanorbel (approved):
🐳

PR #1297, opened 2026-04-14
Fix instrumented tests assertion

sdsantos:
Hopefully this is enough. They are all passing locally now, let's see in the CI.


aanorbel (approved):
🐳

PR #1302, opened 2026-04-15
Make instrumented tests wait for bootstrap

sdsantos:
Now that our Initialization runs on app start, and not app open, it has a race condition with our `skipOnboarding`. Now it should wait for BoostrapPreferences to finish at least.


aanorbel (approved):
🐳

PR #1309, opened 2026-04-21
Refactor measurement url and improve locale string

sdsantos:
We were building URLs that could have weird things like `language=ja-` so I ended up refactoring that code and adding tests.


aanorbel (approved):
🐳

PR #1315, opened 2026-04-24
chore: update passport API

aanorbel:


composeApp/src/commonMain/kotlin/org/ooni/passport/models/SubmitError.kt
```diff
@@ -0,0 +1,38 @@
...
+    object DeserializationFailed : SubmitError("deserialization_failed")
+
+    object CredentialError : SubmitError("credential_error")
+
+    data class Unknown(
+        val raw: String,
+    ) : SubmitError(raw)
+
+    companion object {
+        fun fromWire(value: String?): SubmitError? =
```
sdsantos: Maybe `fromPassport` is a bit more clear?

composeApp/src/commonMain/kotlin/org/ooni/probe/domain/SubmitMeasurement.kt
```diff
@@ -14,13 +16,13 @@ import org.ooni.probe.shared.monitoring.Instrumentation
...
     private val submitMeasurementWithUser: suspend (
         String,
-        MeasurementModel.ReportId,
     ) -> Result<ResponseData, Throwable?>,
     private val engineSubmit: suspend (String) -> Result<SubmitMeasurementResults, MkException>,
     private val readFile: ReadFile,
     private val deleteFiles: DeleteFiles,
     private val updateMeasurement: suspend (MeasurementModel) -> Unit,
     private val deleteMeasurementById: suspend (MeasurementModel.Id) -> Unit,
+    private val handleSubmitOutcome: suspend (VerificationStatus, SubmitError?) -> Unit = { _, _ -> },
```
sdsantos: Why the default lambda?

composeApp/src/commonMain/kotlin/org/ooni/probe/domain/credentials/HandleSubmitOutcome.kt
```diff
@@ -0,0 +1,42 @@
...
+                operation = "SubmitOutcomeError",
+                data = mapOf(
+                    "verification_status" to verificationStatus.name,
+                    "error" to error.code,
+                ),
+            )
+        }
+
+        when (error) {
+            SubmitError.ManifestNotFound -> retrieveManifest()
```
sdsantos: If we failed to already fetch a Manifest, and we try on every background run, I don't think we should retry on every failed submission with ManifestNotFound too.

composeApp/src/commonMain/kotlin/org/ooni/probe/di/Dependencies.kt
```diff
@@ -655,17 +659,38 @@ class Dependencies(
...
+        )
+    }
+    private val clearCredential by lazy {
+        ClearCredential(deleteSecureStorage = secureStorage::delete)
+    }
+    private val handleSubmitOutcome by lazy {
+        HandleSubmitOutcome(
+            retrieveManifest = { retrieveManifest() },
+            clearCredential = clearCredential,
+            signalUpdateRequired = {},
```
sdsantos, line 674: We should mark this as TODO and create a ticket for it, if it's not done yet.

composeApp/src/commonMain/kotlin/org/ooni/probe/domain/credentials/StampMeasurement.kt
```diff
@@ -0,0 +1,44 @@
...
+
+class StampMeasurement(
+    private val passportGetProbeId: PassportGetProbeId,
+    private val getCredential: suspend () -> Credential?,
+    private val json: Json,
+) {
+    suspend operator fun invoke(content: String): String {
+        val credential = getCredential() ?: return content
+        val parsed = try {
+            json.parseToJsonElement(content) as? JsonObject
```
sdsantos: You could use `as` instead, and catch the CastException below, to avoid it failing silently.

composeApp/src/commonMain/kotlin/org/ooni/probe/domain/credentials/ResolveSubmissionPolicy.kt
```diff
@@ -0,0 +1,34 @@
...
+    )
+
+    operator fun invoke(
+        manifest: Manifest,
+        probeCc: String,
+        probeAsn: String,
+    ): Ranges? {
+        for (entry in manifest.manifest.submissionPolicy) {
+            val ccOk = entry.match.probeCc == WILDCARD || entry.match.probeCc == probeCc
+            val asnOk = entry.match.probeAsn == WILDCARD || entry.match.probeAsn == probeAsn
```
sdsantos: We don't need to worry about case sensitivity for CC and ASNs?

composeApp/src/commonMain/kotlin/org/ooni/probe/domain/credentials/SubmitMeasurementWithUser.kt
```diff
@@ -60,6 +70,25 @@ class SubmitMeasurementWithUser(
...
 
+    private fun buildCredentialConfig(
+        manifest: Manifest,
+        credential: Credential?,
+        data: MeasurementData,
+    ): SubmitCredentialConfig? {
+        if (credential == null) return null
+        val ranges = resolveSubmissionPolicy(manifest, data.probeCc, data.probeAsn) ?: run {
+            Logger.w("No submission_policy entry for cc=${data.probeCc} asn=${data.probeAsn}")
+            return null
```
sdsantos, line 99: Without `ageRange` and `measurementCountRange` we shouldn't send any credential then? Just to confirm, could not find that in the documentation.

sdsantos (approved):
Should we create a ticket for storing the verification status of a measurement? Maybe to show it in the UI in the future?

PR #1330, opened 2026-05-04
Upgrade to Java 25

### LINKED ISSUE #1323: desktop: NullPointerException with SceneState.update()
Crash report affecting multiple users: https://ooni.sentry.io/issues/7388689860/events/320f662cd2994c239e2450b05a368049/

sdsantos:
Closes #1323 

Bug fixed on JavaFX 26: https://bugs.openjdk.org/browse/JDK-8334593

Android build complains with `Kotlin does not yet support 25 JDK target, falling back to Kotlin JVM_24 JVM target`. But the Kotlin version 2.3 already supports Java 25. It may be the Kotlin version that's embedded in Gradle 9.3, which is still 2.2. 🤷 

But it runs fine on both Android and Desktop.

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -4,7 +4,7 @@ description: 'Setup Java, checkout and setup gradle'
 inputs:
   java_version:
     required: false
-    default: '17'
+    default: '25'
 
 runs:
   using: "composite"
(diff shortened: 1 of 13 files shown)


aanorbel (approved):
🐳

PR #1350, opened 2026-05-18
Avoid duplicate desktop update error reports

sdsantos:
Double Logger.w/e means 2 different reports on Sentry. We just need one. And Appcast URL was already logged earlier.

Screenshot 2026-05-18 at 11 23 12


aanorbel (approved):
🐳

PR #1359, opened 2026-05-21
feat: update passport dependency and refactor measurement count policy

aanorbel:


composeApp/src/commonMain/kotlin/org/ooni/probe/domain/SubmitMeasurement.kt
```diff
@@ -64,6 +64,7 @@ class SubmitMeasurement(
                     uid = result.value.uid,
                 )
                 updateMeasurement(newMeasurement)
+                Logger.w { "Measurement Submission successful: ${newMeasurement.uid}" }
```
sdsantos: We should not trigger a warning for a successful submission, otherwise Sentry is going to get flooded with this.

sdsantos (approved):
But we need to remove that log.

composeApp/src/commonMain/kotlin/org/ooni/probe/domain/SubmitMeasurement.kt
aanorbel: This was a way to get the measurement uid since the measurements currently show as not uploaded.

PR #1361, opened 2026-05-22
feat: use build directory for desktop debug data and cache

### LINKED ISSUE #1353: Different data/cache folders for desktop debug builds
To avoid data overlap, the debug and production builds should not share the same data and cache folders.

aanorbel:
Closes #1353


sdsantos (approved):
Working great

PR #1362, opened 2026-05-22
feat: add support for Microsoft Store screenshot capture

aanorbel:


sdsantos (approved):
But we should avoid so many screenshots inside the repo, specially the ones we don't use for publishing with the apps.

PR #1371, opened 2026-05-28
feat: add manual refresh button to Descriptors screen

### LINKED ISSUE #1365: desktop: Top-level descriptors update button
Right now, on desktop, you only see the "Check for updates" button if you have installed one run link. But since we now can check for updates for OONI descriptors as well, it should be a top-level button. I suggest a Topbar Icon button, with a refresh icon. This is not needed for mobile, since users can pull to refresh.

aanorbel:
Closes #1365

|.|.|
|-|-|
|Screenshot 2026-06-01 at 15 44 28 | Screenshot 2026-06-01 at 15 44 35 |


sdsantos (commented):
It works well 👍 I would put the refresh between the Search and the Add icons, but it's an unimportant detail.

But I forgot to say that we need to remove the "Check for Updates" button we also have inside the installed descriptors section:

Image

sdsantos (commented):
Still has the "Check for Updates" button on the Descriptors screen.

sdsantos (approved):

Issue #1341, opened 2026-05-11
Results fail to update sometimes


sdsantos:
Occasionally, the results screen gets stuck and does not update with the latest run results. Only an app restart can recover then. Both me and @aanorbel have experienced it at least one time, but fail to reproduce it consistently.

Possibilities:
- Results aggregation into runs is leaving the new results out: `RunListItem.aggregateResults`
- `ResultRepository.countByFilter` somehow stops working

PR #1372, opened 2026-05-29
fix(desktop): pin database access to a single dedicated thread

aanorbel:
Related to https://github.com/ooni/probe-multiplatform/issues/1341


sdsantos (approved):
Good catch

PR #1375, opened 2026-05-29
feat: handle report_id in measurement submission response

aanorbel:


sdsantos (commented):
In the interest of eventually dropping the `report_id` support, we should update our measurements checks to account either for `report_id` or `measurement_uid`.

sdsantos (approved):

PR #1376, opened 2026-05-30
feat: make JavaFX optional for desktop distributions

aanorbel:


composeApp/build.gradle.kts
```diff
@@ -176,7 +197,7 @@ android {
         targetSdk = libs.versions.android.targetSdk
             .get()
             .toInt()
-        versionCode = 300 // Always increment by 10. See fdroid flavor below
+        versionCode = 302 // Always increment by 10. See fdroid flavor below
```
sdsantos: Why bump this? To do another distribution release in the mac app store? If so, we can always bump to 310 anyway, so there's no confusion.

composeApp/src/desktopNoJavaFxMain/kotlin/org/ooni/probe/ui/shared/OoniWebView.desktop.kt
```diff
@@ -0,0 +1,143 @@
...
+ * opens the (allow-listed) URL in the user's system browser via the same
+ * [openUrl] path used elsewhere on desktop, and the screen shows a short
+ * notice with a button to re-open the link.
+ *
+ * This file references no `javafx.*` symbols, so the App Store classpath
+ * never loads a JavaFX class — avoiding `NoClassDefFoundError` even though the
+ * JavaFX jars are absent.
+ */
+@Composable
+actual fun OoniWebView(
```
sdsantos, line 17: Why not use the `Dependencies.isWebViewAvailable()` for opening in an external browser? Instead of having a separate OoniWebView implementation and having to worry about flavors?

sdsantos (approved):

PR #1387, opened 2026-06-08
feat: add verification status and update required prompts

aanorbel:


composeApp/src/commonMain/kotlin/org/ooni/probe/di/Dependencies.kt
```diff
@@ -745,6 +745,8 @@ class Dependencies(
 
     // ViewModels
 
+    fun launchUpdateAction(): Boolean = OrganizationConfig.installUrl?.let { launchAction(PlatformAction.OpenUrl(it)) } ?: false
```
sdsantos, line 748: Ideally the installUrl for Android installed from Google Play would be the actual google play link. But it's likely not easy for us to distinguish that.

sdsantos (approved):

sdsantos (approved):

PR #1388, opened 2026-06-08
Don't keep XML instance

sdsantos:
The Android Studio profiler said this was using 500KB of 8MB of our heap space. So I tried no longer caching it, and the app is now using only 6.2MB. This is also something we don't use all the time, so I feel this is fine like this.


aanorbel (approved):
🐳

PR #1390, opened 2026-06-08
Improve article images memory size

sdsantos:
Improve article images heap size by using the size constraint recommendation here: https://coil-kt.github.io/coil/compose/#rememberasyncimagepainter


aanorbel (approved):
🐳

PR #1405, opened 2026-06-15
Avoid an empty string with the duration is zero

sdsantos:
Our dashboard was showing "Estimated time left: " without a duration in front when it rounded to zero. It felt weird, I think it's better to show `0s`.


aanorbel (approved):
🐳

PR #1424, opened 2026-06-22
Filter DB updates to minimize writes and triggers

sdsantos:
The most important is the `markAllAsDone` query that runs the most often and the P90 is 122ms.


aanorbel (approved):
🐳

PR #1434, opened 2026-06-26
feat(sql): add indexed query for counting unviewed completed results

### LINKED ISSUE #1422: Optimize countAllNotViewed SQL query
We had some warnings to it taking more than 1 second for [some users](https://ooni.sentry.io/issues/7456269329/events/dab5cfaf2d5f4adcba77bcba678dc276/?project=4508325642764288&query=installerStore%3A%5Bcom.android.vending%2Ccom.huawei.appmarket%5D&referrer=previous-event&sort=freq). Sentry is saying the median is low (~6ms) but the 90 percentile is 100ms, which is not ideal for a query that is running all the time (because it's used for the bottom bar).

We should check if we can just query the Result table without merging with the Measurement table. Or if we're missing a valuable index.

aanorbel:
Closes https://github.com/ooni/probe-multiplatform/issues/1422

```pre
config                                   |   median |      p90
OLD  LEFT JOIN + COUNT(DISTINCT)         |    8.014 |    8.237
NEW  EXISTS (no new indexes)             |    0.744 |    0.845
NEW  + Result(is_done,is_viewed)         |    0.543 |    0.588
NEW  + Measurement(result_id,is_done)    |    0.307 |    0.335
NEW  + both.                             |    0.182 |    0.208
```

### DIFF
--- a/composeApp/src/commonMain/sqldelight/org/ooni/probe/data/Result.sq
+++ b/composeApp/src/commonMain/sqldelight/org/ooni/probe/data/Result.sq
@@ -19,6 +19,7 @@ CREATE TABLE Result(
 CREATE INDEX idx_result_start_time ON Result (start_time);
 CREATE INDEX idx_result_descriptor_name ON Result (descriptor_name);
 CREATE INDEX idx_result_task_origin ON Result (task_origin);
+CREATE INDEX idx_result_is_done_is_viewed ON Result (is_done, is_viewed);
 
 CREATE VIEW ResultWithNetworkAndAggregates AS
 SELECT *,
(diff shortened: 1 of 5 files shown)


composeApp/src/commonMain/sqldelight/migrations/19.sqm
```diff
@@ -0,0 +1,2 @@
+CREATE INDEX idx_result_not_viewed ON Result (is_done, is_viewed)
```
sdsantos: Using a composite index is an interesting approach (is_done + is_viewed). But we have to note that it would only be used when we're filtering by is_done, or is_done + is_viewed. It will not be used if we're filtering by is_viewed only.

composeApp/src/commonMain/sqldelight/migrations/19.sqm
```diff
@@ -0,0 +1,2 @@
...
+WHERE is_done = 1 AND is_viewed = 0;
```
sdsantos: Why use a WHERE here? This means that the index won't be used when we're just filtering by `is_done` and a bunch of other cases. Seems quite limiting...

composeApp/src/commonMain/sqldelight/org/ooni/probe/data/Result.sq
```diff
@@ -100,12 +102,15 @@ WHERE (
 ) AND Result.is_viewed = 0;
 
 countAllNotViewed:
-SELECT COUNT(DISTINCT Result.id)
+SELECT COUNT(*)
```
sdsantos, line 104: Let's make sure we measure if this change is faster than before.

sdsantos (approved):
Nice work, thorough benchmark and I can reproduce the findings.

Do you want to keep the benchmark, `@Ignore` it, or can we delete it?

PR #1436, opened 2026-06-26
feat(settings): add in-app language selection support for desktop

### LINKED ISSUE #1425: desktop: Support switching languages
A user asked by email about the ability to switch app languages on desktop.

aanorbel:
Closes https://github.com/ooni/probe-multiplatform/issues/1425

### DIFF
--- a/androidApp/src/main/kotlin/org/ooni/probe/AndroidApplication.kt
+++ b/androidApp/src/main/kotlin/org/ooni/probe/AndroidApplication.kt
@@ -39,6 +39,7 @@ import org.ooni.probe.config.ProxyConfig
 import org.ooni.probe.data.models.BatteryState
 import org.ooni.probe.data.models.PlatformAction
 import org.ooni.probe.di.Dependencies
+import org.ooni.probe.shared.LanguageSupport
 import org.ooni.probe.shared.Platform
 import org.ooni.probe.shared.PlatformInfo
 import java.util.Locale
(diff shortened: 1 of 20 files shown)


composeApp/src/commonMain/kotlin/org/ooni/probe/shared/PlatformInfo.kt
```diff
@@ -10,6 +10,7 @@ data class PlatformInfo(
     val knownBatteryState: Boolean = true,
     val knownNetworkType: Boolean = true,
     val supportsInAppLanguage: Boolean = false,
+    val managesLanguageInApp: Boolean = false,
```
sdsantos: Maybe we can merge these 2 with an enum? Or maybe even take language settings out of the PlatformAction and join them all?

sdsantos (commented):
The supportedLanguageCodes list is not the ideal one for this language list. It does not have the region codes. That means the language picker is not working for Portuguese or Chinese for example.

androidApp/src/main/kotlin/org/ooni/probe/AndroidApplication.kt
```diff
@@ -101,7 +102,11 @@ class AndroidApplication : Application() {
             osVersion = Build.VERSION.SDK_INT.toString(),
             model = "${Build.MANUFACTURER} ${Build.MODEL}",
             requestNotificationsPermission = Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU,
-            supportsInAppLanguage = Build.VERSION.SDK_INT >= 33,
+            languageSupport = if (Build.VERSION.SDK_INT >= 33) {
+                LanguageSupport.SYSTEM_SETTINGS
+            } else {
+                LanguageSupport.NONE
```
sdsantos, line 108: Why not use the new interface now on Android < 33?

sdsantos (approved):

PR #1450, opened 2026-07-06
feat: improve error reporting for measurement submissions

aanorbel:


composeApp/src/commonMain/kotlin/org/ooni/probe/domain/credentials/SubmitMeasurementWithUser.kt
```diff
@@ -45,12 +46,29 @@ class SubmitMeasurementWithUser(
                 probeAsn = data.probeAsn,
                 credentialConfig = credentialConfig,
             ).onSuccess { result ->
+                val outcome = result.decodeSubmitOutcome(json)
```
sdsantos, line 49: It's ok to access `bodyText` on an error response right?

sdsantos (approved):

composeApp/src/commonMain/kotlin/org/ooni/probe/domain/credentials/SubmitMeasurementWithUser.kt
aanorbel: I dont think there are any secrets in the error body, but @DecFox and @LDiazN  will know better

PR #1453, opened 2026-07-06
Android in-app languague picker

### LINKED ISSUE #1452: android: In-App language change for Android < 33

sdsantos:
Closes #1452 

I wasn't able to finish it yet, but you can continue @aanorbel if you have time.

Missing:
- Make the DesktopLocaleController apply to all platforms
- I'm using the `DesktopBuildConfig` for all other platforms too. We should either rename it, or move SUPPORTED_LANGUAGES to a different generic BuildConfig file.

What I've done:
- Fixed a bug: we were setting the Android's `languageSupport` the wrong way around
- `getLanguageNameByCode` for Android
- Renamed `supportedLanguageCodes` to `descriptorLanguageCodes` because it should only be used for the run links localization, and made it match the locales we get from the API. We were mixing app languages with those.


composeApp/src/iosMain/kotlin/org/ooni/probe/SetupDependencies.kt
```diff
@@ -83,15 +84,15 @@ class SetupDependencies(
...
      * override AppleLanguages to the first supported language from the user's
      * preferred languages, or English as a last resort.
      * This must run before any Compose code so that Locale.current and
-     * Compose Resources resoalve to a supported language.
+     * Compose Resources resolve to a supported language.
      */
     private fun ensureSupportedLocale() {
         val preferred = NSLocale.preferredLanguages.mapNotNull { it as? String }
-        val supported = OrganizationConfig.supportedLanguageCodes
+        val supported = DesktopBuildConfig.SUPPORTED_LANGUAGES.map { it.split("-", "_").first() }
```
sdsantos, line 95: It still feels a bit weird that something called `DesktopBuildConfig` is being used on all platforms.

sdsantos (commented):
Working great 👍

composeApp/src/iosMain/kotlin/org/ooni/probe/SetupDependencies.kt
aanorbel: checkout https://github.com/ooni/probe-multiplatform/pull/1455/changes#diff-6c124e0b54178e14b9e1518c3bc78945544b54781cae48cd5c6f02375cd81674R150-R168

sdsantos: > checkout https://github.com/ooni/probe-multiplatform/pull/1455/changes#diff-6c124e0b54178e14b9e1518c3bc78945544b54781cae48cd5c6f02375cd81674R150-R168

Just saw it now 👍

PR #1455, opened 2026-07-07
feat(passport): bump Passport to `0.1.4-alpha` and support proxy/timeouts

### LINKED ISSUE #1454: feat(passport): update passport to `0.1.4-alpha`
Update to  https://github.com/ooni/ooniprobe-rs/releases/tag/v0.1.4-alpha/ and add support for timeout and proxy

aanorbel:
Closes https://github.com/ooni/probe-multiplatform/issues/1454

### DIFF
--- a/composeApp/src/commonMain/kotlin/org/ooni/passport/PassportBridge.kt
+++ b/composeApp/src/commonMain/kotlin/org/ooni/passport/PassportBridge.kt
@@ -17,18 +17,24 @@ interface PassportBridge :
         url: String,
         headers: List<KeyValue>,
         query: List<KeyValue>,
+        proxy: String?,
+        timeout: Float?,
     ): Result<PassportHttpResponse, PassportException>
 
     override fun post(
         url: String,
         headers: List<KeyValue>,
         payload: String,
+        proxy: String?,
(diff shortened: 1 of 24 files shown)


composeApp/src/commonMain/kotlin/org/ooni/passport/PassportBridge.kt
```diff
@@ -17,25 +17,33 @@ interface PassportBridge :
         url: String,
         headers: List<KeyValue>,
         query: List<KeyValue>,
+        proxy: String?,
+        timeout: Float?,
```
sdsantos, line 21: I'm thinking if the PassportBridge implementations should be the ones fetching the proxy and the timeout, instead of delegating that to callers... 🤔

sdsantos (approved):

composeApp/src/commonMain/kotlin/org/ooni/passport/PassportBridge.kt
aanorbel: I would agree with you. `proxy`, `timeout` , `query`, `headers` should all be optional params, `PassportBridge` would get them on init or something like that

PR #1459, opened 2026-07-09
chore(engine,passport): depricate engine.httpDo in favor of `PassportGet`

### LINKED ISSUE #1458: chore(engine,passport): depricate `engine.httpDo` in favor of `PassportGet`

aanorbel:
Closes https://github.com/ooni/probe-multiplatform/issues/1458

### DIFF
--- a/composeApp/src/commonMain/kotlin/org/ooni/probe/di/Dependencies.kt
+++ b/composeApp/src/commonMain/kotlin/org/ooni/probe/di/Dependencies.kt
@@ -18,9 +18,12 @@ import org.ooni.engine.NetworkTypeFinder
 import org.ooni.engine.OonimkallBridge
 import org.ooni.engine.SecureStorage
 import org.ooni.engine.TaskEventMapper
+import org.ooni.engine.models.Result
 import org.ooni.passport.PassportBridge
+import org.ooni.passport.models.PassportException
+import org.ooni.passport.models.PassportHttpResponse
 import org.ooni.probe.Database
-import org.ooni.probe.DesktopBuildConfig
+import org.ooni.probe.SharedBuildConfig
 import org.ooni.probe.background.RunBackgroundTask
 import org.ooni.probe.config.BatteryOptimization
 import org.ooni.probe.config.FlavorConfigInterface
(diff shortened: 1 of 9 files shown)


composeApp/src/commonMain/kotlin/org/ooni/probe/domain/descriptors/FetchDescriptor.kt
```diff
@@ -1,36 +1,63 @@
...
+import org.ooni.passport.models.PassportException
+import org.ooni.passport.models.PassportHttpResponse
 import org.ooni.probe.config.OrganizationConfig
 import org.ooni.probe.data.models.Descriptor
+import org.ooni.probe.data.models.ProxyOption
+import org.ooni.probe.domain.credentials.CredentialsConstants
 
 class FetchDescriptor(
-    private val engineHttpDo: suspend (method: String, url: String, taskOrigin: TaskOrigin) -> Result<String?, MkException>,
+    private val passportGet: (
```
sdsantos, line 23: Use the `PassportGet` interface to avoid having to specify all arguments + return here.

composeApp/src/commonMain/kotlin/org/ooni/probe/di/Dependencies.kt
```diff
@@ -191,6 +192,19 @@ class Dependencies(
     @VisibleForTesting
     var backgroundContext: CoroutineContext = Dispatchers.IO
 
+    // Passport
+    // The HTTP GET function the migrated fetch use cases call (descriptor links,
+    // articles, GeoIP release checks, proxy health). Exposed as a swappable var so
+    // instrumented tests can stub it directly without a fake PassportBridge.
+    @VisibleForTesting
```
sdsantos, line 201: A test implementation of the `PassportBridge` could also achieve this goal, like we did for the Engine.

sdsantos (approved):
Here it's also clear that if the PassportBridge was the one fetching the proxy, it would simplify our code a lot. We could even do a similar approach to the Engine+OonimkallBridge, which is having a middleware classe between the PassportBridge and the classes using it, called Passport or PassportEngine, or something like that.

composeApp/src/commonMain/kotlin/org/ooni/probe/domain/descriptors/FetchDescriptor.kt
aanorbel: I will spin up a new PR to fix this now

PR #1472, opened 2026-07-16
feat(submission): handle unparseable reports gracefully and improve atomic file writes

### LINKED ISSUE #1451: NMS: Fix stamp measurement issue
Error report: https://ooni.sentry.io/issues/7590414485/?project=4508325642764288&query=%21is%3Aarchived%20is%3Aunresolved%20installerStore%3A%5Bcom.android.vending%2Ccom.huawei.appmarket%5D%20release%3A%226.1.1%20%28350%29%22%20%21geo.country_code%3A%5BRU%2CCM%5D&referrer=issue-stream&sort=freq

aanorbel:
Closes #1451


composeApp/src/commonMain/kotlin/org/ooni/probe/data/disk/WriteFile.kt
```diff
@@ -26,22 +29,57 @@ class WriteFileOkio(
...
+            }
+
+            // Atomic replace: write to a sibling temp file in the SAME directory, then rename it
+            // onto the final path. A process killed mid-write can only leave a stale `.tmp`, never a
+            // truncated `absolutePath`. (See the corrupt-measurement-report investigation.)
+            val tmp = (parent ?: absolutePath).resolve("${absolutePath.name}.${Uuid.generateV4()}.tmp")
+
+            // Defensive: delete any orphaned `.tmp` siblings left by prior killed writes.
+            parent?.let { dir ->
+                runCatching {
```
sdsantos, line 47: Maybe we don't care about logging errors here very much, but I'm wondering if we shouldn't logging them anyway...

composeApp/src/commonMain/kotlin/org/ooni/probe/data/disk/WriteFile.kt
```diff
@@ -26,22 +29,57 @@ class WriteFileOkio(
...
             return
         }
 
         try {
-            fileSystem
-                .run { if (append) appendingSink(absolutePath) else sink(absolutePath) }
-                .use { sink ->
-                    sink.buffer().use {
-                        it.writeUtf8(contents)
+            if (append) {
```
sdsantos: This makes me think if we shouldn't have another AppendFile class just for the append use case, since it's so different now.

composeApp/src/commonMain/kotlin/org/ooni/probe/data/disk/WriteFile.kt
```diff
@@ -26,22 +29,57 @@ class WriteFileOkio(
...
+                // Never regress to "cannot write at all" if a filesystem can't do an atomic move.
+                Logger.w("Atomic write failed for $path, falling back to direct write", e)
+                usedAtomic = false
+                runCatching { fileSystem.delete(tmp, mustExist = false) }
+                fileSystem.sink(absolutePath).use { sink ->
+                    sink.buffer().use { it.writeUtf8(contents) }
+                }
+            }
+
+            Instrumentation.reportTransaction(
```
sdsantos: This will mean 1 transaction every 5 seconds on every device, due to the logs write. Not sure we want that many transactions going on...

composeApp/src/commonMain/kotlin/org/ooni/probe/data/disk/WriteFile.kt
```diff
@@ -26,22 +29,57 @@ class WriteFileOkio(
...
                 }
+            }
+
+            var usedAtomic = true
+            try {
+                fileSystem.write(tmp) { writeUtf8(contents) }
+                fileSystem.atomicMove(tmp, absolutePath)
+            } catch (e: IOException) {
+                // Never regress to "cannot write at all" if a filesystem can't do an atomic move.
+                Logger.w("Atomic write failed for $path, falling back to direct write", e)
```
sdsantos: Do all platforms support atomic moves? Otherwise this warning log will get sent every 5 seconds on those platforms.

composeApp/src/commonMain/sqldelight/org/ooni/probe/data/Measurement.sq
```diff
@@ -72,6 +72,7 @@ SELECT * FROM Measurement
 WHERE Measurement.is_done = 1
 AND Measurement.is_uploaded = 0
 AND (:filterByResultId = 0 OR Measurement.result_id = :resultId)
+AND (Measurement.is_upload_failed = 0 OR Measurement.upload_failure_msg IS NULL OR Measurement.upload_failure_msg NOT LIKE 'Report unparseable:%')
```
sdsantos: This query isn't used much, so it doesn't matter to have this extra condition here. 

But I don't like that `Report unparseable` here is hardcoded, and needs to match the other constant in the code. Plus this is very easy to filter in memory.

Not a big deal, we can leave it as it is. But I wanted to give it an heads up.

sdsantos (approved):
I'm wondering if we could use an easier approach for unparsable reports. Why not just mark those measurements as `is_failed=true`? I know the engine did not report them as failed, but if we end up with an unparsable report, it's like it failed anyway, we can't do nothing with it. And that way it would appear to users as a measurement that failed to upload, and they can do nothing about it.

composeApp/src/commonMain/kotlin/org/ooni/probe/data/disk/WriteFile.kt
aanorbel: i would remove

composeApp/src/commonMain/sqldelight/org/ooni/probe/data/Measurement.sq
aanorbel: i get the concern, is_failed would be better

sdsantos (approved):

PR #1480, opened 2026-07-21
feat(credentials): add anonymous credentials management UI and reset functionality

### LINKED ISSUE #1479: anon-cred: add explainer screen and credential reset

aanorbel:
Closes: https://github.com/ooni/probe-multiplatform/issues/1479

|.|.|
|-|-|
| Screenshot_20260727_122114 |  Screenshot_20260727_201044 |
| Screenshot_20260727_201054 | . |


sdsantos (approved):
Looks nice. Did you confirm with the rest of the team that we are ready to release something like this? Just asking because of what @agrabeli said on Slack.

Regarding the UX:
a) I think we could make it more clear what are the consequences of resetting the credentials. Like no data lost, your new measurements with be associated with a new probe ID, something like that... Otherwise you have no idea why do it.
b) I feel after a reset we could try to do a new registration straight away. What do you think?

agrabeli:
Thanks for preparing this PR, it's very exciting to see the Anonymous Credentials in the OONI Probe apps! I have shared minor copy edits privately in a spreadsheet.

PR #1481, opened 2026-07-22
feat(network, articles, workers, submission): enhance offline handling and retry mechanisms

aanorbel:


composeApp/src/commonMain/kotlin/org/ooni/probe/domain/articles/RefreshArticles.kt
```diff
@@ -6,65 +6,115 @@ import kotlinx.coroutines.awaitAll
...
+        try {
+            refresh(skipIntervalCheck)
+        } finally {
+            inFlight.unlock()
+        }
+    }
+
+    private suspend fun refresh(skipIntervalCheck: Boolean) {
+        if (!isOnline()) {
+            // Expected, not a bug: no stamp of any kind, so the refresh runs as soon as a network
```
sdsantos, line 54: What does "no stamp of any kind" mean? And expected to be offline?

sdsantos (approved):

composeApp/src/commonMain/kotlin/org/ooni/probe/domain/articles/RefreshArticles.kt
aanorbel: here we are checking if the device is offline first. 

we dont throw an exception or anything and we know failure is the expected outcome.

PR #1500, opened 2026-08-03
Fix TestProxy in async. Add progress indicator to Preparing.

sdsantos:
Due to the dev back-end bug with the `/health` endpoint, I realized that our TestProxy was actually running in sync, and not async like we wanted it to.

I also added a indeterminate progress indicator to the "Preparing" state, because check-in can take a bit, and that way it looks more clear that the app is already doing things.


aanorbel (approved):
🐳

PR #1505, opened 2026-08-07
chore: update iOS deployment target to `15.0`

aanorbel:


sdsantos (approved):

sdsantos (commented):
There's no easy way to have all these values shared in a single constant right?

aanorbel:
I can experiment with picking the value from a configuration file.

sdsantos (approved):

PR #1506, opened 2026-08-07
feat: split out reusable `probe-core`

aanorbel:


probeCore/src/commonMain/kotlin/org/ooni/engine/models/TestType.kt
```diff
@@ -73,16 +41,16 @@ sealed class TestType {
         override val isBackgroundRunEnabled: Boolean = false,
         override val isManualRunEnabled: Boolean = true,
     ) : TestType() {
-        override val labelRes: StringResource = Res.string.Test_Experimental_Fullname
-        override val iconRes: DrawableResource = Res.drawable.test_experimental
+        override val labelResKey: String = "Test_Experimental_Fullname"
+        override val iconResKey: String = "test_experimental"
```
sdsantos: I feel we're not gaining much from this String keys. We can't use `when` on them. Why not remove them and just have a regular `when` on the extension function?

probeCore/src/commonMain/kotlin/org/ooni/engine/Engine.kt
```diff
@@ -172,14 +172,19 @@ class Engine(
         MAX_RUNTIME_DISABLED
     }
 
+    private fun buildSoftwareName(taskOrigin: TaskOrigin): String {
```
sdsantos: Why do we need to repeat this here?

probeCore/src/commonMain/kotlin/org/ooni/probe/data/models/ScaledValue.kt
```diff
@@ -33,11 +27,11 @@ data class ScaledValue(
...
 
-    val unitStringId
+    val unitStringKey: String
         get() = when (unit) {
-            Unit.KB -> Res.string.TestResults_Kbps
-            Unit.MB -> Res.string.TestResults_Mbps
-            Unit.GB -> Res.string.TestResults_Gbps
+            Unit.KB -> "TestResults_Kbps"
+            Unit.MB -> "TestResults_Mbps"
+            Unit.GB -> "TestResults_Gbps"
```
sdsantos: Not gaining much here too, with this String keys.

sdsantos:
In general it looks good. I feel, with more work, can improve a bit more things. But that's not necessary for a Proof of Concept. I'm thinking of:
- probeCore could be a plain kotlin module, instead of a Kotlin Multiplatform module
- probeCore wouldn't have CLI code in it (like those Gateways)
- Dependencies would be split between CoreDependencies (networkTypeFinder, engine/passport bridges...) and ComposeDependencies (isWebViewAvailable, supportedLanguageTags...)
- There may be need for a shared desktop module with the autorun and update code for both the desktop compose app and the desktop cli app.

PR #1511, opened 2026-08-10
Avoid triple reporting issues with retrieving manifest

sdsantos:
When the API call to retrieve a manifest failed, we were reporting it 3 times to Sentry. We just need the report inside `RetrieveManifest`.


aanorbel (approved):
🐳

Issue #1512, opened 2026-08-10
Avoid parsing a submission response that we're not supposed to handle


sdsantos:
Right now, we sometimes get a 502 response from the server with an un-parsable response. See sentry reports: https://ooni.sentry.io/issues/7657671308/

We're always trying to parse the response we get, whatever the HTTP status code. We should confirm with the back-end what are the status codes we're supposed to parse: 200? 400? 500? 502 is unlikely.

PR #1521, opened 2026-08-17
fix: json parse error for non-successful request

aanorbel:
Related to https://github.com/ooni/probe-multiplatform/issues/1512


sdsantos (approved):
The changes make sense, I'm just wondering if we're missing some scenarios where the measurement is rejected with 400 or something that's not a 502, and we still get a verificationStatus we should handle.

PR #1525, opened 2026-08-17
Avoid duplicated warning reports on user registration

sdsantos:
We're already reporting from inside `CredentialResponse` and `SetCredential`.


aanorbel (approved):
🐳

PR #1532, opened 2026-08-17
Fix results missing after midnight

### LINKED ISSUE #1531: Results after midnight do not show unless the app is re-opened
Reported by an user via email.

sdsantos:
Closes https://github.com/ooni/probe-multiplatform/issues/1531

The lambda ensures that new dates are created whenever we create a new filter (every time the `ResultsViewModel` is instantiated). The `MAX_DATE` covers the scenario where the app stays opened on the `ResultsScreen/ViewModel` for a long time.


aanorbel (approved):
🐳

PR #1533, opened 2026-08-18
feat: add support for private-dns

### LINKED ISSUE #1530: mobile: Handle private DNS settings
Related with https://github.com/ooni/probe/issues/2857

We should detect if the private DNS setting is enabled and:
- Warn the user if they are testing with it enabled (similar to what we do with VPNs)
- Add that annotation to the measurements submitted

aanorbel:
Closes https://github.com/ooni/probe-multiplatform/issues/1530

### DIFF
--- a/composeApp/src/commonMain/kotlin/org/ooni/engine/DefaultResolverTypeFinder.kt
+++ b/composeApp/src/commonMain/kotlin/org/ooni/engine/DefaultResolverTypeFinder.kt
@@ -1,0 +1,21 @@
+package org.ooni.engine
+
+import org.ooni.engine.models.NetworkType
+import org.ooni.engine.models.ResolverType
+
+/**
+ * Default implementation of [ResolverTypeFinder] that delegates to a [NetworkTypeFinder].
+ * - Returns [ResolverType.VPN] when network type is VPN.
+ * - Returns [ResolverType.Unknown] when network type is [NetworkType.NoInternet].
+ * - Returns [ResolverType.System] for all other active network types (e.g. Wi-Fi, Mobile, Ethernet).
+ */
+class DefaultResolverTypeFinder(
+    private val networkTypeFinder: NetworkTypeFinder,
+) : ResolverTypeFinder {
+    override fun invoke(): ResolverType =
+        when (val networkType = networkTypeFinder()) {
+            is NetworkType.VPN -> ResolverType.VPN
+            is NetworkType.NoInternet -> ResolverType.Unknown
+            else -> ResolverType.System
+        }
+}
--- a/composeApp/src/commonMain/kotlin/org/ooni/engine/DefaultResolverTypeFinder.kt
+++ b/composeApp/src/commonMain/kotlin/org/ooni/engine/DefaultResolverTypeFinder.kt
@@ -7,7 +7,7 @@ import org.ooni.engine.models.ResolverType
  * Default implementation of [ResolverTypeFinder] that delegates to a [NetworkTypeFinder].
  * - Returns [ResolverType.VPN] when network type is VPN.
  * - Returns [ResolverType.Unknown] when network type is [NetworkType.NoInternet].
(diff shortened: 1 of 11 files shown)


composeApp/src/commonMain/kotlin/org/ooni/engine/Engine.kt
```diff
@@ -150,6 +151,7 @@ class Engine(
         ),
         annotations = TaskSettings.Annotations(
             networkType = networkTypeFinder(),
+            privateDns = privateDnsFinder(),
```
hellais: We should change this to be `resolver_type` that can take values of:
* `private_dns`
* `vpn`
* `system` (when `private_dns IS FALSE AND vpn IS FALSE`)
* `unknown` (for all others)

hellais: We should then also update the ooni/spec to include this as an annotation: https://github.com/ooni/spec/blob/master/data-formats/df-000-base.md#annotations

composeApp/src/commonMain/kotlin/org/ooni/engine/DefaultResolverTypeFinder.kt
```diff
@@ -0,0 +1,21 @@
...
+ * - Returns [ResolverType.System] for all other active network types (e.g. Wi-Fi, Mobile, Ethernet).
+ */
+class DefaultResolverTypeFinder(
+    private val networkTypeFinder: NetworkTypeFinder,
+) : ResolverTypeFinder {
+    override fun invoke(): ResolverType =
+        when (val networkType = networkTypeFinder()) {
+            is NetworkType.VPN -> ResolverType.VPN
+            is NetworkType.NoInternet -> ResolverType.Unknown
+            else -> ResolverType.System
```
sdsantos: Did you confirm with @hellais that we should default to System instead of Unknown on iOS and Desktop? The spec says: `system: when Private DNS is false and VPN is false.`

sdsantos (approved):

composeApp/src/commonMain/kotlin/org/ooni/engine/DefaultResolverTypeFinder.kt
aanorbel: This is the general assumption for the case of non-android platform.  @hellais , any comments here.

hellais: Thanks for flagging this. Yes, the default should be `unknown` if we aren't able to establish if they are indeed using an encrypted resolver. Since the spec states that both conditions need to be false and in this case the Private DNS condition is undefined, so we should go into the `unknown` branch. It might be worth making this even more clear in the spec to avoid confusion (eg. say that on iOS and desktop we don't detect the use of an encrypted DNS and hence in this case we say private).

It might be worth doing some further research into this as follow up work and depending on what turns out we could convert this to the `system` string also on `iOS` and `macOS`. I did some very cursory testing of this based on what I read in here: [https://github.com/paulmillr/encrypted-dns](<https://github.com/paulmillr/encrypted-dns>) and it seems like encrypted DNS profiles only apply to certain applications, but not to others (eg. dig, littlesnitch, etc.). We should verify if that is indeed the case for OONI Probe as well and if so we can perhaps consider detection unnecessary on these platforms. I did test however that if I load those settings, then web browsers will indeed use encrypted DNS on macOS too.

aanorbel: I will put some more time into checking this out.

sdsantos (approved):

composeApp/src/commonMain/kotlin/org/ooni/engine/DefaultResolverTypeFinder.kt
aanorbel: This will be tracked in the linked issue https://github.com/ooni/probe-multiplatform/issues/1575

PR #1534, opened 2026-08-18
fix(passport): improve error handling and registration retries

aanorbel:


composeApp/src/commonMain/kotlin/org/ooni/passport/models/PassportException.kt
```diff
@@ -19,6 +19,16 @@ sealed class PassportException(
         message: String?,
     ) : PassportException(message)
 
+    class HttpStatus(
```
sdsantos: `HttpStatus` does not sound like an error/exception. `HttpRequestUnsuccessful` or something similar is more fitting.

composeApp/src/commonMain/kotlin/org/ooni/probe/domain/SubmitMeasurement.kt
```diff
@@ -124,6 +122,24 @@ class SubmitMeasurement(
...
 
+    /**
+     * The OONI submission contract guarantees that a `2xx` response stored the measurement and
+     * that `4xx` and `5xx` responses did not. Client errors therefore stop the legacy fallback,
+     * except for the transient [408](https://www.rfc-editor.org/rfc/rfc9110#section-15.5.9) and
+     * [429](https://www.rfc-editor.org/rfc/rfc6585#section-4) responses.
+     *
+     * See <https://github.com/ooni/backend/blob/e5aa51a275622dec67aea763c41f7eded33cbcbd/ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py#L908-L910>.
+     */
+    private fun shouldFallbackToLegacy(reason: Throwable?): Boolean =
```
sdsantos: So we fallback to legacy if there's a bug/issue in the server (5xx responses), but we don't fallback if there's something wrong with the registration, or it's outdated (like a 400)? And if the user never recovers from that state, we never upload their submissions?

sdsantos (commented):
The code looks good. I'm just concern in what scenarios we should not fallback to the legacy submission. I'm worried there are scenarios where submissions can get stuck between always failing with passport, but never falling back to legacy.

PR #1537, opened 2026-08-20
feat: include `X-Protocol-Version` header in `PassportHttpClient` requests

aanorbel:
Related to OOPRIV-67

### DIFF
--- a/composeApp/src/commonMain/kotlin/org/ooni/probe/di/Dependencies.kt
+++ b/composeApp/src/commonMain/kotlin/org/ooni/probe/di/Dependencies.kt
@@ -9,7 +9,6 @@ import androidx.datastore.preferences.core.Preferences
 import app.cash.sqldelight.db.SqlDriver
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.IO
-import kotlinx.coroutines.flow.map
 import kotlinx.serialization.json.Json
 import okio.FileSystem
 import okio.Path.Companion.toPath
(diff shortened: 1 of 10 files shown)


DecFox (approved):
LGTM!

LDiazN (approved):
LGTM!

composeApp/src/commonMain/kotlin/org/ooni/probe/di/Dependencies.kt
```diff
@@ -618,6 +619,7 @@ class Dependencies(
             passportAuthRegister = passportBridge::userAuthRegister,
             passportAuthSubmit = passportBridge::userAuthSubmit,
             getProxyOption = proxyManager::selected,
+            getProtocolVersion = { getManifest().map { it?.meta?.protocolVersion } },
```
LDiazN: Is this getting the protocol version from the manifest? I think the one we want to deliver here is the protocol version currently running in the app, example of how this is done in the Python bindings: [https://github.com/ooni/userauth/blob/cae34f0813f171fb941ac9514e088402d4cb276e/ooniauth-py/src/protocol.rs#L42-L47](<https://github.com/ooni/userauth/blob/cae34f0813f171fb941ac9514e088402d4cb276e/ooniauth-py/src/protocol.rs#L42-L47>)

It probably has to be exposed by the passport library but @mehul can confirm this. If not exposed we can assume "0.1.0" as that's the only version there is now

The idea is to get the protocol version running on the app to route them to different versions of the backend in case they are running a soon-to-deprecate version

The manifest is not a good proxy for this because the version there is what was reported by the backend and it can get updated

LDiazN (approved):

DecFox (approved):
LGTM!

sdsantos (approved):
Looks good. But it feels weird asking the passport library for a value, and then giving it back to the library in the form of an HTTP header. Couldn't the library just do it automatically? What do you think @DecFox?

DecFox:
The library already adds the `X-Protocol-Version` header and overrides any incoming values with the version referenced by the library: https://github.com/ooni/ooniprobe-rs/blob/main/ooniprobe-ffi/src/userauth.rs#L23, for the register and submit functions. I think this was meant as a sanity check to ensure the apps always send the header on their end as well. We can get rid of it if we think it does not help us.

PR #1544, opened 2026-08-26
fix: handle ASN "AS0" as invalid in measurement

### LINKED ISSUE #1522: Do not test and submit measurements with an ASN = 0
This is generating 400 response errors when submitting measurements:

https://ooni.sentry.io/issues/7657671610/?project=4508325642764288&query=%21is%3Aarchived%20is%3Aunresolved%20installerStore%3A%5Bcom.android.vending%2Ccom.huawei.appmarket%5D%20release%3Aorg.openobservatory.ooniprobe%406.2.0%2B360&referrer=issue-stream&sort=freq

But we need to log the issue to Sentry anyway, even if we abort the tests.

aanorbel:
Closes https://github.com/ooni/probe-multiplatform/issues/1522

### DIFF
--- a/composeApp/src/commonMain/kotlin/org/ooni/engine/models/TaskEvent.kt
+++ b/composeApp/src/commonMain/kotlin/org/ooni/engine/models/TaskEvent.kt
@@ -17,7 +17,10 @@ sealed interface TaskEvent {
         val countryCode: String?,
         val geoIpdb: String?,
         val networkType: NetworkType,
-    ) : TaskEvent
+        val shouldAbort: Boolean = false,
+    ) : TaskEvent {
+        override fun shouldAbort(): Boolean = shouldAbort
+    }
 
     data class Log(
         val level: String?,
@@ -75,6 +78,8 @@ sealed interface TaskEvent {
     data class TaskTerminated(
         val index: Int,
     ) : TaskEvent
+
+    fun shouldAbort(): Boolean = false
 
     fun isCancelled() =
         when (this) {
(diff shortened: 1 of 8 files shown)


composeApp/src/commonMain/kotlin/org/ooni/engine/models/TaskEvent.kt
```diff
@@ -17,7 +17,10 @@ sealed interface TaskEvent {
         val countryCode: String?,
         val geoIpdb: String?,
         val networkType: NetworkType,
-    ) : TaskEvent
+        val shouldAbort: Boolean = false,
+    ) : TaskEvent {
+        override fun shouldAbort(): Boolean = shouldAbort
```
sdsantos, line 22: You don't need the `val`, you can just do:

```
override fun shouldAbort(): Boolean = probeAsn.isAsnZero()
```

composeApp/src/commonMain/kotlin/org/ooni/probe/domain/SubmitMeasurement.kt
```diff
@@ -79,6 +82,22 @@ class SubmitMeasurement(
             return marked
         }
 
+        if (reportProbeAsn(report).isAsnZero()) {
```
sdsantos, line 85: I feel this could be joined with the `reportStructuralError` and turned into a check for report errors. Both try to parse the report. But not a big deal, this works as well, just feel there's a big of duplicated code, and we might do more checks in the future.

sdsantos (approved):

PR #1569, opened 2026-09-10
feat(oonirun): implement descriptor creation

### LINKED ISSUE #1486: Create Run Links inside the app

aanorbel:
Closes https://github.com/ooni/probe-multiplatform/issues/1486 , https://github.com/ooni/run/issues/97


composeApp/src/commonMain/kotlin/org/ooni/probe/domain/auth/DecodeAuthResponse.kt
```diff
@@ -0,0 +1,25 @@
+package org.ooni.probe.domain.auth
+
+import kotlinx.serialization.SerializationException
+import kotlinx.serialization.json.Json
+import org.ooni.engine.models.Failure
+import org.ooni.engine.models.Result
+import org.ooni.engine.models.Success
+import org.ooni.passport.models.PassportHttpResponse
+
+internal inline fun <reified T> decodeAuthResponse(
```
sdsantos, line 10: This is a more functional approach to decomposing the domain use case. We usually create a class `DecodeAuthResponse` with `Json` as a dependency passed in the constructor, and then a single invoke method.

I noticed you also did this for `AuthSessionMapping.kt`, but there you choose to use an extension function, a third different approach. I feel we should be consistent, although there's nothing wrong with either approach.

composeApp/src/commonMain/kotlin/org/ooni/probe/domain/auth/RefreshSession.kt
```diff
@@ -0,0 +1,42 @@
...
+import org.ooni.engine.models.Failure
+import org.ooni.engine.models.Result
+import org.ooni.engine.models.Success
+import org.ooni.passport.PassportBridge
+import org.ooni.passport.models.PassportException
+import org.ooni.passport.models.PassportHttpResponse
+import org.ooni.probe.config.OrganizationConfig
+import org.ooni.probe.data.models.AuthSession
+
+class RefreshSession(
```
sdsantos, line 14: This isn't used yet. Should it be called automatically on app start?

composeApp/src/commonMain/kotlin/org/ooni/probe/domain/auth/AuthException.kt
```diff
@@ -0,0 +1,23 @@
...
+    data class Network(
+        override val cause: PassportException,
+    ) : AuthException(cause.message)
+
+    data class Decode(
+        override val cause: Throwable,
+    ) : AuthException(cause.message)
+
+    /** No session is stored locally, so there is nothing to refresh or attach as a Bearer token. */
+    data object NotLoggedIn : AuthException("Not logged in")
```
sdsantos, line 22: Android Studio is complaining that exceptions should not be objects.

composeApp/src/commonMain/kotlin/org/ooni/probe/domain/descriptors/CreateDescriptor.kt
```diff
@@ -0,0 +1,94 @@
...
+        url: String,
+        payload: String,
+        extraHeaders: List<PassportBridge.KeyValue>,
+    ) -> Result<PassportHttpResponse, PassportException>,
+    private val getStoredSession: suspend () -> AuthSession?,
+    private val json: Json,
+) {
+    suspend operator fun invoke(request: OONIRunLinkCreateRequest): Result<Descriptor, OonirunApiError> {
+        val session = getStoredSession() ?: return Failure(OonirunApiError.Unauthorized("Not logged in"))
+        if (request.author != session.emailAddress) {
```
sdsantos, line 38: You could also just override it.

composeApp/src/commonMain/kotlin/org/ooni/probe/domain/descriptors/CreateDescriptor.kt
```diff
@@ -0,0 +1,94 @@
...
+                OonirunApiError.BadRequest(
+                    "email_address must match the email address of the user who created the oonirun link",
+                ),
+            )
+        }
+
+        // Fields left unset (name_intl, icon, expiration_date, ...) must be omitted rather than
+        // sent as explicit nulls - the server only applies its own defaults (e.g. expiration_date
+        // defaulting to now+6 months) when a field is absent, not when it's present as null.
+        val body = json.encodeToString(
```
sdsantos, line 49: Can't we create a `Json` object that has the specific configuration to not encode nulls and use it here, so we can just do `val body = json.encodeToString(request)`?

composeApp/src/commonMain/kotlin/org/ooni/probe/domain/descriptors/ParseOonirunApiError.kt
```diff
@@ -0,0 +1,46 @@
...
+import kotlinx.serialization.json.JsonObject
+import kotlinx.serialization.json.JsonPrimitive
+import kotlinx.serialization.json.contentOrNull
+import kotlinx.serialization.json.jsonPrimitive
+
+/**
+ * Maps a non-2xx oonirun/ooniauth response to a typed [OonirunApiError], since the service does
+ * not return a uniform error body shape (see [OonirunApiError] for the shapes handled).
+ */
+fun parseOonirunApiError(
```
sdsantos, line 14: Again we used to use classes for this. It's fine it now we're using a different pattern. But it's important to agree when do we prefer to do this, versus using a class.

composeApp/src/commonMain/kotlin/org/ooni/probe/ui/descriptor/create/CreateDescriptorViewModel.kt
```diff
@@ -0,0 +1,453 @@
...
+
+    private sealed interface OptionsResult {
+        data class Valid(
+            val value: JsonObject?,
+        ) : OptionsResult
+
+        data object Invalid : OptionsResult
+    }
+
+    data class State(
```
sdsantos, line 342: Instead of using State with a Phase inside, it's more correct to make State a sealed interface, to avoid impossible states like a `Phase.LoggedOut` with a `session`.

composeApp/src/commonMain/kotlin/org/ooni/probe/ui/descriptor/create/CreateDescriptorViewModel.kt
```diff
@@ -0,0 +1,453 @@
...
+import org.ooni.engine.models.Result
+import org.ooni.engine.models.Success
+import org.ooni.engine.models.TestType
+import org.ooni.probe.data.models.AuthSession
+import org.ooni.probe.data.models.Descriptor
+import org.ooni.probe.data.models.NetTest
+import org.ooni.probe.domain.auth.AuthException
+import org.ooni.probe.domain.descriptors.OonirunApiError
+
+class CreateDescriptorViewModel(
```
sdsantos, line 28: This ViewModel is already quite big. It wouldn't hurt to turn the Login steps into a separate screen. But it's just a suggestion.

sdsantos (changes requested):
In terms of UI, here are my suggestions:

1. Creating should install the link straight away
1. I think we only need the web connectivity links for now, like on the web
1. The list of urls could work as in the Choose Websites (maybe next we can even offer that option also from inside the Choose Websites screen: run once, or create link to run more times and share with others)
1. The FAB button needs to explicitly say create or create web test (and maybe the Add button needs to be something more than just a plus icon)
1. The "Log out" button could be just a text link after "Logged in.." just so it isn't in the location where we usually place the main action of a screen
1. JSON field is not needed for now
1. Users should be able to set an expiration date, or the default should be longer, like 5 years
1. It would be nice if we could pick the color + icon (but not a requirement for now of course)
1. The app should support the login deep links, so users could login by tapping the link directly
1. The form should validate that there's at least 1 valid URL
1. I don't think we need the share screen at the end. Users can also create links just for themselves, to be able to run them automatically. I think it's easier to just redirect them to the descriptors screen, which has a share button.
1. I don't think the app link is necessary for sharing, the other one works in all circumstances and is more familiar
1. The login screen should have some explanations regarding why do you need to login, and that it's only for this feature, nothing else is tied to your email

composeApp/src/commonMain/kotlin/org/ooni/probe/domain/descriptors/CreateDescriptor.kt
```diff
@@ -0,0 +1,72 @@
...
+    private val getStoredSession: suspend () -> AuthSession?,
+    private val parseOonirunApiError: ParseOonirunApiError,
+    /**
+     * Configured with `explicitNulls = false`: fields left unset (name_intl, icon,
+     * expiration_date, ...) must be omitted rather than sent as explicit nulls, since the server
+     * only applies its own defaults (e.g. expiration_date defaulting to now+6 months) when a
+     * field is absent, not when it's present as null.
+     */
+    private val requestJson: Json,
+    private val responseJson: Json,
```
sdsantos, line 37: We can even sub-class `Json` or wrap it in our own classes, if we need to be super-sure not the mix up the instances.

composeApp/src/commonMain/composeResources/values/strings-common.xml
```diff
@@ -410,9 +410,42 @@
     <string name="DescriptorUpdate_Updates">UPDATES</string>
     <!-- Only used for the desktop version of the app for now -->
     <string name="DescriptorUpdate_CheckUpdates">Check for Updates</string>
+    <string name="Descriptors_CreateFab">Create web test</string>
     <string name="AddDescriptor_AutoRunDisabled">Auto-run is disabled. Please enable it to run tests automatically.</string>
     <string name="Descriptor_LastTestResult">Last Test Result</string>
 
+    <string name="CreateDescriptor_Title">Create New Link</string>
+    <string name="CreateDescriptor_Login_Email">Email address</string>
+    <string name="CreateDescriptor_Login_Explanation">Logging in is only needed to create and manage OONI Run links. Your email address isn\'t used for anything else in the app.</string>
```
sdsantos: Avoid using `\'` and use an apostrophe `’`. You can see in the screenshot that the \ is visible there.

androidApp/src/main/kotlin/org/ooni/probe/MainActivity.kt
```diff
@@ -96,6 +96,16 @@ class MainActivity : ComponentActivity() {
...
 
+            "login" -> {
+                val token = uri.getQueryParameter("token")
+                if (token.isNullOrBlank()) {
+                    deepLinkFlow.tryEmit(DeepLink.Error)
+                    Logger.e { "Login deep link without a token: $uri" }
+                } else {
+                    deepLinkFlow.tryEmit(DeepLink.Login(token))
+                }
+            }
```
sdsantos, line 107: My email comes with a `run.test.ooni.org/login` link and not a `ooni://` link.

sdsantos: And the insert token screen should now also mention you can just click the link in the email.

sdsantos (approved):
There's also some UI improvements I want to do in the form section. I'll try to do them myself today.

sdsantos:
How it looks after my changes. Main things I did:
- I tried to fit more things in less height
- Maybe the icon and expiration buttons more integrated with the form
- Use Create Web Test everywhere
- Make the main buttons stand out more
- Fixed the descriptors screen bottom padding, so the FAB does not hide any content.

Screenshot_20260921_181320

PR #1573, opened 2026-09-14
chore: migrate to `sentryKmp` plugin and improve crash reporting configurations

aanorbel:


androidApp/build.gradle.kts
```diff
@@ -5,7 +5,7 @@ plugins {
     alias(libs.plugins.androidApplication)
     alias(libs.plugins.jetbrainsCompose)
     alias(libs.plugins.jetbrainsComposeCompiler)
-    alias(libs.plugins.sentry)
+    alias(libs.plugins.sentryAndroid)
```
sdsantos, line 8: Maybe we can drop this now?

sdsantos (approved):

androidApp/build.gradle.kts
aanorbel: This is still required. Check out the samples https://github.com/getsentry/sentry-kotlin-multiplatform/blob/main/sentry-samples/kmp-app-spm/androidApp/build.gradle.kts#L6

PR #1578, opened 2026-09-17
chore: lower macOS minimum system version to `12.0.0`

### LINKED ISSUE #1380: desktop: Investigate if we can lower the MacOS min version
The current minimum version (`14.8.3`) is leaving some users behind. We don't know how many for sure, because we don't have any analytics into that. But we should document what's keeping us from lowering the minimum version, and see if we can if we can address them.

aanorbel:
Closes https://github.com/ooni/probe-multiplatform/issues/1380

Screenshot 2026-09-17 at 11 52 52


sdsantos (approved):
Wow, I didn't expect it was just that. Not sure why we went with the higher version 🤷

aanorbel:
> Wow, I didn't expect it was just that. Not sure why we went with the higher version 🤷

Can't remember why we upgraded too 😀

